PR #7, opened 2021-01-03
Add darktheme prototype: WIP

JojiiOfficial:
I created a dark-theme implementation prototype. I couldn't come up with an idea how to implement the dark theme for the menu buttons properly, since they are constants and can't be made variables easily. So this is WIP for now. I thought about making dark-theme a feature but tbh I prefer it to be an option which should become a preference in some kind of option/config menu later on.

### DIFF
--- a/psst-gui/src/ui/theme.rs
+++ b/psst-gui/src/ui/theme.rs
@@ -9,13 +9,17 @@ pub fn grid(m: f64) -> f64 {
 pub const GRID: f64 = 8.0;
 
 pub const WHITE: Color = Color::WHITE;
+pub const DARK_WHITE: Color = Color::rgb8(34, 40, 49);
+pub const DARK_SELECTED: Color = Color::rgb8(57, 62, 70);
 pub const BLACK: Color = Color::BLACK;
 pub const GREY_1: Color = Color::grey8(0x33);
 pub const GREY_2: Color = Color::grey8(0x4f);
+pub const DARK_GREY_2: Color = Color::grey8(0xef);
 pub const GREY_3: Color = Color::grey8(0x82);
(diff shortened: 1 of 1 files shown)


jpochyla:
Thanks! Real theming support would indeed require:

- Refactoring all current use of constants from `ui/theme.rs` to env keys.
- Figuring out all the colors etc.
- Implementing custom widget that switches between the themes in a performant manner (i.e. similar to `EnvScope` but caching the env and updating it only when the theme selector from `State` changes).
- Adding it to the config dialog.
- Bonus points: implementing OS-theme retrieval support to `druid-shell`, ideally with some "theme changed" event.

florianjr:
I saw a gist on how to support automatic event-based dark mode switching on macOS. I'll try it tomorrow and write here again :)

JojiiOfficial:
But on MacOS it would not be cross platform. Or what exactly are you talking about?

thibautvdv:
For later on: I prefer using a config file, like [alacritty](https://github.com/alacritty/alacritty) does with YAML, or [Starship](https://github.com/starship/starship) with TOML. This gives more flexibility than environment variables IMO. Just my two cents!

JojiiOfficial:
Yeah a config file sounds good... But for this, as @jpochyla already said, quite some things need to be done first.

jpochyla:
As of 408e8abd47d0933ca1a47c1aebb5c299b940acfe there is a simple dark theme added. More work on colors is needed, it's not very pleasant right now, but it's a start. I'm leaving this issue open, as reacting to (at least Mac) OS theme changes would be really nice to have as well.

Screenshot 2021-01-14 at 19 13 05

PR #19, opened 2021-01-12
actions: initial build action for ubuntu

gokayokyay:
Here's a basic build action for ubuntu, uses cache action too which speeds up things a lot :smile:


jpochyla:
Thank you!

PR #25, opened 2021-01-20
[WIP] Random and Loop controls

reynn:
First pass at implementing random and looping to the project.

this is really rough right now and currently is only configurable in code. I have some ideas on where to go from here but would like some feedback.

1. remove the QueueBehavior struct from queue and instead have the next/previous methods of queue accept the QueueBehavior instead, then having the behavior be part of the PlayerDelegate in the GUI.
2. PlayerDelegate has a player_queue but it is currently a vec, this could potentially be changed to the Queue struct in core/audio_player.

I'm sure I could have missed stuff since I am still learning the code base, if there is a better way you think this could be accomplished I'm happy to hear them :)


jpochyla:
Thank you! I think the general approach looks good.

> remove the QueueBehavior struct from queue and instead have the next/previous methods of queue accept the QueueBehavior instead, then having the behavior be part of the PlayerDelegate in the GUI.

`PlayerDelegate` is on a separate thread and is communicating with player through events, so this wouldn't fly I think, as the player needs to have access to the queue beh. on it's own (i.e. when the track ends). The solution to that would indeed be what you propose:

> PlayerDelegate has a player_queue but it is currently a vec, this could potentially be changed to the Queue struct in core/audio_player.

...basically moving the queue management out of `AudioPlayer`, together with `Previous`, `Next`, `LoadQueue`, and changing how the `Preload` command works. I think this would simplify things a bit, but is quite some work, and I think we can do this later in a separate PR.

So what remains now is:
- a command to set a queue behaviour to the player,
- a bit of cleanup in the queue,
- the icons :)

I will have time next week to take a look at this and get it to merge. Thanks again!

jpochyla:
Hi, I've merged the PR with two changes: 2cc178546620201eef708306797bed3e3848db34 and d4b74a29ab8123b7dd448ebb31e16281d52ad72b. Please try it, we can make changes if something is wrong. Thanks!

PR #27, opened 2021-01-22
actions: Combine duplicate code and add Windows support

Kesefon:
I set it up to continue on cargo test errors, because some minivorbis related tests fail on Windows without causing any noticeable problems later.


jpochyla:
Thanks!

Issue #16, opened 2021-01-07
OS Bundles


gokayokyay:
Hello!
First of all, thanks for the amazing project. :heart:  I really appreciate for your efforts to save us from lots of cpu usage (no offense electron's awesome too just a little heavy).
When do you plan to create OS specific bundles? Compilation of the release profile is way too slow. Takes up to an hour with a 1 gig linux instance.

jpochyla:
Hi! Thanks, that's great to hear. Platform bundles are definitely on the roadmap, once we have a CI (GitHub actions?) working.

gokayokyay:
Hmm okay then I'll fork and try some stuff with actions, but won't promise anything :smile:

jpochyla:
We now have Ubuntu and Mac CI working, and the artifacts (just naked executables for now) are available at the run page.

KorbsStudio:
I mean, I guessing seeing a DEB, AppImage, Flatpak, or maybe a Snap file would be nice.

TheEvilSkeleton:
Just created a Flatpak manifest: https://github.com/jpochyla/psst/pull/141. I'm awaiting for approval so it creates a Flatpak bundle.

reesericci:
AppImages anyone?

cedws:
I want to create a Scoop manifest but the project would need to adopt Releases for this to be viable. Well, that's not entirely true, but pulling a "nightly" binary in a package is not really any good.

Any possibility of this?

jpochyla:
I was planning to start releasing once we reach a certain feature set, most probably including Connect support, playlist manipulation, and once we solve the audio-related problems (device switching, audio buffer underruns on high CPU). I should reify these somehow as a Milestone here.

reesericci:
Flatpak seems to be pulling away in the Linux packaging sphere. I now believe that Flatpak is the best option for Psst.

KorbsStudio:
Flatpak is the best met, meaning pushing it it Flathub will be a good idea.

TheEvilSkeleton:
We could directly package it to Flathub. That's what OBS Studio did https://github.com/obsproject/obs-studio/pull/5569.

SO9010:
@jpochyla is the still an issue as we have got the downloads section?

Peque:
@jacksongoode It seems that https://github.com/jpochyla/psst/issues/170 was closed as duplicate of this issue (https://github.com/jpochyla/psst/issues/16).

[...]

PR #28, opened 2021-01-24
actions: Build deb package

Kesefon:
This builds a Debian (.deb) package for easier installation on Debian/Ubuntu based systems, as suggested in #16.
This PR is currently based on PR #27, but I can easily rebase it onto the current master branch if you want.


jpochyla:
Thank you!

PR #31, opened 2021-02-01
fix looping error

reynn:
The current implementation will not play the last song while looking songs. This should resolve the problem and allow all tracks to play.


jpochyla:
Thanks!

Issue #52, opened 2021-05-25
Can't type credentials


ShawkMusic:
XFCE Manjaro Linux 5.10.36-2-MANJARO

When I open the program, it shows settings and asks for my credentials. I can click on the buttons and on the text fields but when I try to type, nothing gets written, and whatever I type gets typed on whatever the window was that I had opened before opening psst. 

If I click save and go to the main app and try to type there (eg the search bar), typing works fine, but when i open settings again, it doesnt send the text to it, it only sends it to the main app. This means it's probably a problem with the settings specifically.

This is pretty crucial for me because I can't sign in at all.

KaMiila:
I am having the same issue on Artix (open-rc port of Arch linux).

jpochyla:
Seems like an issue in Linux Druid key handling.

KaMiila:
if i could figure out the syntax for username and password in the config.json file I could just input it manually in there.  but I have no idea what it should look like.

{
  "credentials": null,
  "audio_quality": "High",
  "theme": "Dark"
}

jpochyla:
The credentials field contains a token retrieved from Spotify, it's unfortunately not possible to put username & password there.

Tangential: Psst really needs somebody to take care of Linux and Windows platforms, I've created #54 for this.

KaMiila:
What generates or pulls in that token?  If I can figure out what my token is I can just input it directly to the config.json file

Liberatys:
@jpochyla @KaMiila I'm looking into the issue with username/password entry on linux

jpochyla:
@Liberatys I've pushed an update that uses the latest Druid master. If it doesn't help, let's report the issue upstream.

Liberatys:
@jpochyla The changes in dependencies did not resolve the issue 🤔 I found a way to make the inputs work on Linux and also show the window on the proper screen if you are using multiple screens. https://github.com/jpochyla/psst/pull/55. I don't know in what way that influences the appearance on macOS.

freidrichen:
I was able to work around the problem by entering my username/password in the config like this:

```
"credentials": {
  "username": "my-username",
  "auth_data": "my-password",
  "auth_type": 0
}
```

jpochyla:
> I don't know in what way that influences the appearance on macOS.

Only in very minor way, let's keep it like this.

> I was able to work around the problem by entering my username/password in the config like this:

Oh, great idea! I totally forgot the type is an enum.

[...]

PR #55, opened 2021-05-27
Remove modal level

Liberatys:
This is related to #52. By removing the modal level the window can get the proper focus on linux.

A side effect is also that the window will be shown on the currently focused screen if you have multiple screens connected in linux (tested on arch)


KaMiila:
This was my issue it now logs me in but I have another issue now so I will open a new ticket.

jpochyla:
Thanks a lot! <3

PR #56, opened 2021-05-27
Implement initial volume slider

Liberatys:
This pull requests implements a working version of a volume slider. It was one of the pain points when using the application :D. As I was unsure about the location, I just put it somewhere where it could be easily accessed and thus not intrude on the rest of the application.

The slider allows values between 0 - 100 and displays the values as integer in the label (but is chosen as f32).

<details>
<summary>Image of the sidebar</summary>

screen

</details>

I so far had no issues with it and it seems to perform well.

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -31,6 +31,7 @@ pub struct PlaybackController {
     thread: Option<JoinHandle<()>>,
     output_thread: Option<JoinHandle<()>>,
     media_controls: Option<MediaControls>,
+    volume: f64,
 }
 
 impl PlaybackController {
@@ -40,6 +41,7 @@ impl PlaybackController {
             thread: None,
             output_thread: None,
             media_controls: None,
+            volume: 50.0,
         }
     }
 
@@ -85,8 +87,8 @@ impl PlaybackController {
                     Self::handle_media_control_event(event, &sender);
                 }
             })
-            .unwrap();
 
+            .unwrap();
         self.sender.replace(sender);
         self.thread.replace(thread);
         self.output_thread.replace(output_thread);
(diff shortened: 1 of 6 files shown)


psst-gui/src/controller/playback.rs
```diff
@@ -222,6 +224,11 @@ impl PlaybackController {
         self.send(PlayerEvent::Command(PlayerCommand::Seek { position }));
     }
 
+    fn update_volume(&mut self, volume: f64)  {
+        self.send(PlayerEvent::Command(PlayerCommand::SetVolume { volume: ((volume / 100.0) as f32) }));
```
Liberatys, line 228: The slider widget implements Widget<f64> and I then convert it to f32 for miniaudio

Liberatys: Also we have to divide by 100 as the values is chosen in 0 - 100 but miniaudio is expecting 0 - 1

psst-gui/src/controller/playback.rs
```diff
@@ -40,6 +41,7 @@ impl PlaybackController {
             thread: None,
             output_thread: None,
             media_controls: None,
+            volume: 50.0,
```
Liberatys, line 44: I think 50 is a good default?

jpochyla:
Great, thanks!

- I'm not sure about the location. The best we could do, I guess, is a on-hover popup with a vertical slider, somewhere near the playback controls? Hovers are a bit difficult in Druid now, though.
- We don't need to duplicate the `volume` value in the `PlaybackController`, we can diff it in `update`.
- Please please use `rustfmt` for the PRs.

I'll fix the code stuff, we can figure out where to put the slider later.

psst-gui/src/controller/playback.rs
jpochyla: I think that as the default used to be 100, we should keep it.

jpochyla:
I've ended up tweaking the visuals a bit and polishing the code. If you're OK with it we can merge.

Liberatys:
Your solution is much cleaner 👍🏻 nice

PR #62, opened 2021-06-15
Added Apple Silicon support to miniaudio

lonkle:
All this does is add `bindgen` to miniaudio's features which is needed to build this app on Macs powered by Apple Silicon.


jpochyla:
Thank you!

PR #68, opened 2021-07-14
Hide console window on windows

Kethku:
This change will prevent the console window from showing when running release builds on windows


jpochyla:
Great, thank you!

Issue #77, opened 2021-07-15
Add current song title to window title


Kethku:
This would show the song in the taskbar which is very convenient

jpochyla:
Fixed in #78, closing <3

PR #78, opened 2021-07-17
Add song to window title

Kethku:
Solves https://github.com/jpochyla/psst/issues/77


jpochyla:
Great, thanks! It's a bit unfortunate that it allocates on every `AppState` change, that should happen now only every ~100ms or so when a track is playing, or when typing, changing the volume slider etc., but I guess we can live with that (and I think there's no other way to do it now).

Issue #76, opened 2021-07-15
Current song isn't very visible


Kethku:
Its not immediately obvious which song is playing from a list. as far as I can tell, the only think that indicates it is bolding of the dots between song name and song length. Maybe adding a background change and bolding all the text for the current song would be a good way to fix this

Kethku:
Fixed by #80

PR #80, opened 2021-07-17
More visible active track

Kethku:
Addresses https://github.com/jpochyla/psst/issues/76 by adding a new link background state for when the link is active. I added a function to link which takes a predicate and a background color to render when the predicate returns true.

Dark Mode:
image

Light Mode:
image

Note: I noticed that light mode selection theme is a bit too subtle, so I had to boost the alpha for light mode a bit more than I did for dark mode.

### DIFF
--- a/psst-gui/src/main.rs
+++ b/psst-gui/src/main.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 #![allow(clippy::new_without_default)]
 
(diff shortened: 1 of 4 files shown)


jpochyla:
Great, thanks! I'd prefer separating the closure and the active background color, but let's merge this and edit later.

jpochyla:
I've made some changes in 309d66bd8dada97d0ad56e25e53c09b4a39e499b.

PR #97, opened 2021-08-17
Added instructions for RHEL/Fedora

zrthxn:
Added instructions to the README for installing dependencies on RHEL/Fedora. I was trying to get the project working on my machine and figured this would be helpful for others too.
And yes I've tested by building the project and it works.

Also removed the `$` dollar signs before the commands, just to make it easy to use the "copy" button and not have to remove that before pasting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,30 +8,39 @@ Fast Spotify client with native GUI, without Electron, built in Rust. Very early
 
 ##### Linux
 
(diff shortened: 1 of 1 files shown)


jacksongoode:
Yas!

jpochyla:
Thank you!

Issue #101, opened 2021-08-17
PNG album images fail to load


bspammer:
<img width="479" alt="image" src="https://user-images.githubusercontent.com/11244353/129788373-5e3e25dc-adc3-4f37-89c5-1bc0335b10f5.png">

Not all album images are jpegs, e.g. `https://misc.scdn.co/home-mix/family_default.png` is the URL for the above missing image.

bspammer:
Fixed in https://github.com/jpochyla/psst/commit/f47f3e967670656e3e06a0e2cdf6392d92ee9313

PR #102, opened 2021-08-17
Assume PNG images when URL includes ".png"

bspammer:
Fixes https://github.com/jpochyla/psst/issues/101

First time using Rust so not sure if there's a cleaner way to do this :)

### DIFF
--- a/psst-gui/src/delegate.rs
+++ b/psst-gui/src/delegate.rs
@@ -128,8 +128,12 @@ impl Delegate {
                     .unwrap();
             } else {
                 self.spawn(move || {
+                    let image_format = match &location[location.rfind(".").unwrap_or(0)..] {
+                        ".png" => image::ImageFormat::Png,
+                        _ => image::ImageFormat::Jpeg
+                    };
                     let dyn_image = WebApi::global()
-                        .get_image(&location, image::ImageFormat::Jpeg)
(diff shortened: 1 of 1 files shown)


psst-gui/src/delegate.rs
```diff
@@ -128,8 +128,12 @@ impl Delegate {
                     .unwrap();
             } else {
                 self.spawn(move || {
+                    let image_format = match &location[location.rfind(".").unwrap_or(0)..] {
```
Kethku, line 131: Maybe use endswith https://doc.rust-lang.org/std/string/struct.String.html#method.ends_with? Alternatively if its a filepath, extension https://doc.rust-lang.org/std/path/struct.Path.html#method.extension

Kethku: Also rather than using a match for a single case, you can use an `if let` expression https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

jpochyla:
Hi @bspammer! Thanks a lot for the fix, I haven't realized some of the images will be PNGs 🤦‍♂️ I've ended up parsing the returned content-type in f47f3e967670656e3e06a0e2cdf6392d92ee9313.

bspammer:
@jpochyla Cool! Definitely a better way to do it.

Issue #103, opened 2021-08-17
CMD+W doesn't close the window on MacOS


bspammer:

Kethku:
I'm not sure this is a bug. IMHO ctrl-w is reserved for closing tabs or windows for applications which have multiple identical components.

Stooovie:
> I'm not sure this is a bug. IMHO ctrl-w is reserved for closing tabs or windows for applications which have multiple identical components.

that's for windows. in macos, that's been cmd-w since before windows even existed.

Kethku:
In that case, this is an upstream issue I believe. If its truly universal, then it should be handled in druid I think

JuliDi:
This could be fixed by adding a hotkey that sends `command::CLOSE_WINDOW`(https://docs.rs/druid/0.7.0/druid/commands/constant.CLOSE_WINDOW.html). However, this is not intended behaviour for druid on macOS (see https://docs.rs/druid/0.7.0/druid/platform_menus/mac/application/index.html).

While implementing the cmd+h `HIDE_WINDOW` shortcut yesterday, I have also briefly added cmd+w to send `CLOSE_WINDOW`, however, I figured that closing the window doesn't make any sense if you don't want to quit the entire application – the window is gone and so are all your controls. How would you bring back the window once it's closed?
What would be your expected behaviour when pressing cmd+w?

Note: I am not sure why the default menu is overridden (entirely?) by psst since I don't understand the comment in https://github.com/jpochyla/psst/blob/1d5cf324ef4acd7eac1c5bb5b56ca059e494c670/psst-gui/src/ui/menu.rs#L23-L25

henrymoulton:
Cmd + W would close the window but leave music running and the "mini player" active in the top OS menu bar. 

CMD + Q would kill everything.

jpochyla:
This is slightly more difficult than it seems, because i.e. the audio output thread is stored in a `PlaybackController` widget that would get destroyed on closing the window.

JuliDi:
Thanks for the clarification @jpochyla . I gave it a try and realized that closing the window also stops the music.

I have started looking into rust libraries that provide tray icon functionality but didn't yet find anything. Could the `PlaybackController` be held in the tray (if possible) to be more persistent and independent of the window?

peterkos:
[...]

PR #108, opened 2021-08-18
Add cmd+h for hide and ctrl+space for pause/resume

### LINKED ISSUE #91: Addition of keyboard controls
On Mac OS hiding with CMD+H and playing/pausing with space bar etc...

JuliDi:
This should fix #91 in parts, but has a little drawback: you need to press ctrl+space to pause/resume the song.
I could not get the application to react to a normal key even without a modifier (i.e. only pressing space or another key when the psst window is open).
If there is a way to achieve this, I'd be happy to make the according changes.

In case #103 should be implemented (ctrl+w sends druid's `command::CLOSE_WINDOW`), I could add that as well.

### DIFF
--- a/psst-gui/src/cmd.rs
+++ b/psst-gui/src/cmd.rs
@@ -40,7 +40,6 @@ pub const PLAY_TRACKS: Selector<PlaybackPayload> = Selector::new("app.play-track
 pub const PLAY_PREVIOUS: Selector = Selector::new("app.play-previous");
 pub const PLAY_PAUSE: Selector = Selector::new("app.play-pause");
 pub const PLAY_RESUME: Selector = Selector::new("app.play-resume");
-pub const PLAY_PAUSEORRESUME: Selector = Selector::new("app.play-pause_or_resume");
 pub const PLAY_NEXT: Selector = Selector::new("app.play-next");
 pub const PLAY_STOP: Selector = Selector::new("app.play-stop");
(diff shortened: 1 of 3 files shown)


jpochyla:
Hi, thanks for the contribution! While the `Cmd+H` shortcut is definitely Mac-specific, `Ctrl+Space` won't work on other platforms, as we omit the application menu on Windows and Linux. Cross-platform keyboard shortcuts would need to be implemented in a custom controller, handling them in one place, or separately in the specific controllers (i.e. pause/resume in the `PlaybackController`).

JuliDi:
The `Ctrl+Space` shortcut should not be in the application menu, I've added it to the `edit` menu. Isn't that visible on all platforms?
See
https://github.com/jpochyla/psst/blob/b01ac00590fd15d913dc6783f2c9ff33a5c28c0d/psst-gui/src/ui/menu.rs#L9-L15

What do you think about adding a "Playback" menu (like the spotify app) where all those controls are collected and assigned a shortcut? Maybe that'd be a better solution.

> (i.e. pause/resume in the `PlaybackController`)

Could you point me to where that would go in the playback controller? I played around with keydown events in there yesterday but couldn't get anything to work.

JuliDi:
Alright, thanks for your help.

I have reverted the changes regarding pause/resume and left only the `cmd+h` part.

>It would go to the Controller::event method, after the command matching stuff. This might be too cryptic and I will definitely get to it myself very soon, so feel free to omit this from the PR.

Okay, I will have a look and see whether I can make it work, and open a new PR if it does.

jpochyla:
Thank you! ❤️

PR #111, opened 2021-08-18
Add keyboard controls to playback controller

JuliDi:
As discussed in #108, more keyboard controls are added to the playback controller.

For now, this includes:

- `Space` for pause/resume
- `Right/Left Arrow` for next/previous song
- `+/-` for increase/decrease volume

One todo for this change (in a new PR?) would be reading those keys from the config file to make the shortcuts changeable via the settings. But I'd need some input on on how this should be implemented.

### DIFF
--- a/psst-gui/src/cmd.rs
+++ b/psst-gui/src/cmd.rs
@@ -40,7 +40,6 @@ pub const PLAY_TRACKS: Selector<PlaybackPayload> = Selector::new("app.play-track
 pub const PLAY_PREVIOUS: Selector = Selector::new("app.play-previous");
 pub const PLAY_PAUSE: Selector = Selector::new("app.play-pause");
 pub const PLAY_RESUME: Selector = Selector::new("app.play-resume");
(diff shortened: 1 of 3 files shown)


jpochyla:
Thanks! I've re-based on top of master, as it's needed for the focus mechanism to work, and added some code to request focus again after i.e. the search input resigns it. Seems to work!

mattarderne:
Seems like the shortcuts are only recognised when a song is selected. It would be perhaps useful to have `play` shortcut default to the top song on the top playlist of the home section. 

Also macOS "play" shortcut key is ignored in the above circumstance and so itunes loads 

Maybe this is the wrong place for this comment

JuliDi:
Maybe the macOS play shortcut (I guess you are referring to the one on the touch bar?) is not recognized as a media shortcut by druid? I will look into this in the next days. Thank's for bringing this up!

mattarderne:
Yes seems all mac keyboards have a play/pause and back/forward track buttons. 
Loving the app, daily use. Thanks!

PR #120, opened 2021-08-19
Fix wrong expiration time parsing

JuliDi:
The parsing of the expiration time seems to be wrong (probably outdated?), giving me a `failed to parse expiration time from URL ...` warning.

Not sure whether these server responses can differ across operating systems or so. At least building from master on macOS  10.15 I get a response of the form `.../...a35817ca410?__token__=exp=1629466995~hmac=df348...`.

With the changes in this PR, the issue seems fixed – the log says that a new token has been requested because it was outdated.

### DIFF
--- a/psst-core/src/cdn.rs
+++ b/psst-core/src/cdn.rs
@@ -140,13 +140,13 @@ fn parse_total_content_length(response: &ureq::Response) -> u64 {
 }
 
 /// Parses an expiration of an audio file URL.
-/// Expiration is stored at the beginning of the first query parameter, i.e.:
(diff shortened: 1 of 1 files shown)


jpochyla:
Hi @JuliDi, thanks for the fix! It seems like we're getting different URL formats. I've combined the parsing approaches in the latest commit here. Can you please verify it works on your side?

JuliDi:
Very odd... but your commit works on my end, as well: `[2021-08-20T12:18:06Z INFO  psst_core::access_token] access token expired, requesting`

Looks ready to merge.

PR #126, opened 2021-08-20
feat(GUI): Add specific account tab

WaitroseDev:
Not a huge change, but adds a separate account tab instead of leaving everything on the General tab (Looks better imo)


jpochyla:
Thanks for the idea @WaitroseDev! The preferences needed some bigger changes, so I've included separate account tab into 12acdeb.

PR #129, opened 2021-08-20
Store shortcuts in config

JuliDi:
This is a first attempt at storing keyboard shortcuts in the config, following up on  #111.
Ideally, if you open the preferences now, there is a "shortcuts" tab with some default settings. For now, either a single character (like "+" or "-" for the volume controls) or a `Code` as defined by https://docs.rs/keyboard-types/0.6.1/keyboard_types/enum.Code.html can be entered.

At the moment, the entries are only parsed and validated – I couldn't manage to get proper feedback (like an error message or so) back to the UI when the entered shortcut could not be parsed.

It doesn't seem very straightforward to implement those keyboard shortcuts and recording them by the user would be even better (and is on my todo list). Also, there are probably a lot of corner-cases that might be or become problematic. 

So feedback on this is highly appreciated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2003,7 +2003,8 @@ dependencies = [
 [[package]]
 name = "souvlaki"
 version = "0.4.1"
-source = "git+https://github.com/Sinono3/souvlaki#ab184816861bf2d691a1cf30db834cad383f70d6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5834fb7421062570b409e49ecebee68d303866c8ba1c3f81138a82e8eba64081"
 dependencies = [
  "block",
  "cocoa",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,7 +489,7 @@ checksum = "bd0c93bb4b0c6d9b77f4435b0ae98c24d17f1c45b2ff844c6151a07256ca923b"
 [[package]]
 name = "druid"
 version = "0.7.0"
-source = "git+https://github.com/JuliDi/druid?branch=psst#a31679cdc6e0d4bebb96f5c50970ff9afa254851"
+source = "git+https://github.com/jpochyla/druid?branch=psst#30aa0baab6f9801aba16db48793d97a23839856a"
 dependencies = [
  "console_error_panic_hook",
  "druid-derive",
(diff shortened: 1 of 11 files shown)


jpochyla:
Thanks a lot for the PR @JuliDi. Do you think we really need to make the shortcuts configurable? i.e. what other shortcut sets would make sense?

JuliDi:
Making them configurable might prove useful in the future if there are more actions that can have a shortcut (there might be people preferring the F-keys over arrow keys and characters etc). However, this might be my personal preference from using other programs.

Regarding other shortcuts, some ideas that come to mind:

- liking a song
- adding a song to the library
- adding a song to the last used playlist
- play selected song next (this would be good in combination with the ability to use arrow keys to navigate a playlist)
- opening the containing album for the currently playing song
- focus the search bar (i.e. you press "s" and then type the name of a song, no need to click anywhere)

jacksongoode:
Just wondering - it looks like some hotkeys are specified in menu::view_menu but they don't seem to work?

jacksongoode:
Are some hotkeys already specified [here](https://github.com/jpochyla/psst/blob/f6eb1fe8d273af43266407d998aacac87df8c04d/psst-gui/src/ui/menu.rs#L44-L75) - but they don't seem to work (at least on Linux)?

JuliDi:
> Are some hotkeys already specified [here](https://github.com/jpochyla/psst/blob/f6eb1fe8d273af43266407d998aacac87df8c04d/psst-gui/src/ui/menu.rs#L44-L75) - but they don't seem to work (at least on Linux)?

I can confirm that this does not work with Ubuntu 20.04, might be related to this issue: https://github.com/linebender/druid/issues/281
There is no real menu bar as e.g. in macOS and Windows, so that might be an issue.

kataklasm:
I'd love to get configurable hotkeys as well. I'm using my computer without a mouse, just a keyboard. So being able to control Psst without using the mouse would be great. Few recommendations (based on vim control scheme):

focus search bar: /
focus sidebar: h
focus main panel: l
scroll down single line: j
scroll up single line: k
scroll down half page: ctrl-d
scroll up half page: ctrl-u
like song: a

but as mentioned before, these shortcuts should still be configurable by the user. Example usecase: I'm not using one of the usual keyboard layouts, so being able to set my own hotkeys would be great!

JuliDi:
@kataklasm totally agree that there should be some good default bindings and they should be configurable by the user without changing the code and compiling the crate.

As far as I can tell, the issue on Linux with druid is still not resolved.
Maybe there is an alternative solution by now, but I am not really familiar with druid myself. If binding keys to menu entries is not possible on linux, there should be some alternative.
So if anyone has an idea how we could implement this, that would be great.

PR #136, opened 2021-08-22
Added Support for Local Files

sleepychaser:
Parsing of the Spotify configuration for local files and the JSON returned by the web API.

## Changes
Local files imported by Spotify is stored in a custom binary format at `<Spotify` Config `Dir>/Users/<username>-user/local-files.bnk`. When playlist tracks are attempted to be loaded instead of discarding any local tracks attempt to find a local track in the `local-files.bnk` and if a matching title, artist, album, etc. are found then return a properly formatted Track structure

## TODO:
1. Test on platforms other than Linux
2. Attempting to play a local files throws unimplemented
3. Figure out how to decode and play formats other than Vorbis
4. Continue reverse engineering the local-files.bnk so that local files can be added through Psst
5. Maybe look into syncing local files with other devices on the network?


psst-gui/src/data/config.rs
```diff
@@ -157,6 +164,13 @@ impl Config {
             Some,
         )
     }
+
+    pub fn get_username(&self) -> Option<String> {
+        match &self.credentials {
```
jpochyla, line 169: `map` would be better here I think?

psst-gui/src/main.rs
```diff
@@ -33,12 +33,14 @@ fn main() {
     .init();
 
     let config = Config::load().unwrap_or_default();
+    let username = config.get_username();
```
jpochyla, line 36: This won't really work the first time user launches the app. Can we maybe inject the username (or the `LocalTrackManager`) into `WebApi` a bit later?

psst-gui/src/webapi/client.rs
```diff
@@ -26,27 +26,33 @@ use std::{
...
         let agent = default_ureq_agent_builder(proxy_url).unwrap().build();
         Self {
             session,
             agent,
             cache: WebApiCache::new(cache_base),
             token_provider: TokenProvider::new(),
+            local: match username {
+                Some(u) => LocalTrackManager::new(u),
+                _ => None
+            }
```
jpochyla, line 55: Again `map` I personally find preferable here, but maybe we should re-think how the manager is initialized, see above.

psst-gui/src/webapi/client.rs
```diff
@@ -372,7 +378,7 @@ impl WebApi {
 
     // https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-playlist
     pub fn get_playlist(&self, id: &str) -> Result<Playlist, Error> {
-        let request = self.get(format!("v1/playlists/{}", id))?;
+        let request = self.get(format!("v1/me/playlists/{}", id))?;
```
jpochyla, line 381: Is the endpoint really incorrect?

psst-gui/src/webapi/local.rs
```diff
@@ -0,0 +1,400 @@
...
+ * represent.
+ * 
+ * Then a track entry is presented as follows Title -> Artist -> Album -> Local Path -> Trailer.
+ * I have no clue what the trailer represents other than directly after the Local path the bytes
+ * 0x08 -> <varint encoding of track length>, not relevant so far. The title, artist, etc are each
+ * encoded in the following format 0x09(string identifier) -> <varint string size> -> string
+ * If the varint size is 0 this is a null string
+ * 
+ * The end of a trailer for a given track is signified by 0x78, 0x04 from what I can tell
+ */
```
jpochyla, line 70: I really appreciate the docs!

psst-gui/src/webapi/local.rs
```diff
@@ -0,0 +1,400 @@
...
+        Ok(ret)
+    }
+
+    pub fn read_string_with_len(&mut self, len: usize) -> io::Result<String> {
+        let str_bytes = self.read_exact(len)?;
+        match str::from_utf8(&str_bytes) {
+            Ok(s) => Ok(s.to_string()),
+            // TODO: This can definitely be done better
+            Err(_) => Err(io::Error::from(io::ErrorKind::Other))
+        }
```
jpochyla, line 142: Maybe using `String::from_utf8` directly, with `map_err`, would be better?

psst-gui/src/webapi/local.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        return true;
+    }
+
+    pub fn new(username: String) -> Option<Self> {
+        let file_path = match Config::spotify_local_files(username) {
+            Some(p) => p,
+            _ => return None
+        };
```
jpochyla, line 262: `?` should cover this I think?

psst-gui/src/data/track.rs
```diff
@@ -20,6 +20,8 @@ pub struct Track {
     pub track_number: usize,
     pub explicit: bool,
     pub is_local: bool,
+    #[serde(skip_deserializing)]
+    pub local_path: Option<Arc<String>>,
```
jpochyla, line 24: `Arc<str>` would be preferable.

jpochyla (changes requested):
Wow, nice! Added some nitpicks, I'd like to as well do some of my own modifications later, before merging this, if you don't mind.

sleepychaser:
I committed most of the changes you suggested. The problem I was running into was injecting the username into the `WebApi` after log in as it becomes an Arc. I ended up just making the `LocalTrackManager` a singleton similar to `WebApi`. I don't know if this is the best way to do it and there may be a better way with Rust features I'm not familiar with. Feel free to make any changes you see fit.

jpochyla:
Hi @sleepychaser, I've made some more changes (mostly some style nits, functionality should be identical I think). It works on macOS. The playback is next!

jpochyla:
Re. multiple codecs support: this is definitely needed (not even all Spotify tracks are in Vorbis). Options:

1. Get rid of minivorbis, use [Symphonia](https://github.com/pdeljanov/Symphonia) (which seems *excellent*) for the decoding, and miniaudio just for the output.
2. Wait until miniaudio hits a stable version with the highlevel API and vorbis codec included.
3. Keep minivorbis and the MP3 decoder from miniaudio -- I don't like this much.

It's sad the the rust audio output is in such a bad state (rodio can not even keep the buffers full in debug mode, cpal has >100 issues open and maintainers don't seem to care, etc.), so it seems like we're stuck with a C library for this.

jpochyla:
Merged to master, sorry about the delay :) Playback is still not supported, but now that we have Symphonia, it wouldn't be that hard to do.

Kyiro:
Will playback ever be supported?

jpochyla:
Hi @Kyiro, I've opened an issue for this. No ETA, though :(

Issue #146, opened 2021-08-26
FR: Current song shadow not visible


pablodz:
Hello there,

![image](https://user-images.githubusercontent.com/28235457/130884393-2e5b3361-e9bd-48f8-a757-1914b1624971.png)

Please add a shadow or a contrast to see which song of the playlist is current playing because is not visible 

Thanks in advance

PR #148, opened 2021-08-26
Increase opacity of current track overlay in light theme

Kethku:
Addresses https://github.com/jpochyla/psst/issues/146


jpochyla:
Thanks!

Issue #160, opened 2021-08-31
Automatic theme switching


edfloreshz:
I noticed we don't have an **automatic** option for the theme. 

![image](https://user-images.githubusercontent.com/22224438/131546486-93c8480d-c4b6-49d5-8c4f-8eb6ce89a9e5.png)

If nobody's working on this I'd like to do it.

PR #162, opened 2021-08-31
Auto theme switch

edfloreshz:
This PR adds code for detecting the current theme on any OS when selecting the Auto option in settings, it will also be able to detect color scheme changes on the fly.

Solves #160


timgws:
I think you mean it solves #160?

edfloreshz:
Fixed, thanks!

PR #172, opened 2021-09-07
Save/restore queue behavior

aeons:
This PR...

- adds queue behavior to the configuration
- updates and saves the configuration on queue behavior changes
- and restores the queue behavior state on startup

Currently it handles all queue behaviors, but it might make sense to only restore sequential/random (which is my use case), since loop track and loop all are more used on a once-off basis?


jpochyla:
Thank you!

Issue #113, opened 2021-08-18
Add artist view 


panmau:
I like to listen to tracks by a specific artist.
So I suggest adding an artist page that lists all your artists, you can click on one and you will find all saved songs from this artists.

This feature was once in the spotify desktop client but is now gone.. On mobile it's still kinda there but it needs two clicks now.

pythonhacker:
Not sure if there is another issue open for Artist icon, but would be nice to see some Artist artwork (icon/image) in the playing area which is kind of dead empty right now.

PR #176, opened 2021-09-14
gui: Display album art in playback bar

brandly:
Hey, I'm Matt! This app is great. I want to learn more about how it all fits together. I saw [this comment](https://github.com/jpochyla/psst/issues/113#issuecomment-901996044) from @pythonhacker and figured I'd take a stab at it.

My changes here are fairly basic. I saw `cover_widget` functions defined in files like `ui/artist.rs` and followed suit. I thought filling the available space would look nice and somewhat similar to the _Albums_ tab, but an alternative design would be with rounded corners and margins around the image similar to the _Home_ tab. Let me know if you'd like me to change this!

The current code works in most situations. I can play from a playlist:

Screen Shot 2021-09-13 at 9 10 54 PM

From my saved tracks:

Screen Shot 2021-09-13 at 9 11 01 PM

From a Daily Mix:

Screen Shot 2021-09-13 at 9 11 14 PM

But when I play from an album view, no album art is displayed:

Screen Shot 2021-09-13 at 9 11 38 PM

Thanks for taking a look!

### DIFF
--- a/psst-gui/src/data/playback.rs
+++ b/psst-gui/src/data/playback.rs
@@ -49,11 +49,16 @@ pub struct NowPlaying {
 
 impl NowPlaying {
     pub fn cover_image_url(&self, width: f64, height: f64) -> Option<&str> {
-        self.item
-            .album
-            .as_ref()
+        self.item_album()
             .and_then(|album| album.image(width, height))
             .map(|image| image.url.as_ref())
+    }
+
+    pub fn item_album(&self) -> Option<&AlbumLink> {
+        self.item.album.as_ref().or_else(|| match &self.origin {
+            PlaybackOrigin::Album(album) => Some(album),
+            _ => None,
+        })
     }
 }
 
--- a/psst-gui/src/ui/playback.rs
+++ b/psst-gui/src/ui/playback.rs
@@ -16,7 +16,7 @@ use crate::{
         AppState, AudioAnalysis, NowPlaying, Playback, PlaybackOrigin, PlaybackState,
         QueueBehavior, Track,
     },
-    widget::{icons, icons::SvgIcon, Empty, Maybe, MyWidgetExt},
+    widget::{icons, icons::SvgIcon, Empty, Maybe, MyWidgetExt, RemoteImage},
 };
 
 use super::{theme, utils};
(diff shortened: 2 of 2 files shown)


jacksongoode:
Looks great!

jpochyla:
Hi, thanks for the PR. I'd be open to this, but reason the cover image is not there is that the Now Playing section displays the current "playback context" -- album, artist top tracks, or playlist. We could always display the album cover, but:

1. I'm not sure it's always available, i.e. for playlist tracks.
2. If we do it, it should probably act as a separate button, leading to the album view.

We could also display the artist cover for the top tracks context, and the playlist image for the playlist context.

Thoughts?

Kethku:
Just my two cents, but I think having a button that acts differently in various contexts can be more confusing than its worth. Imho its not too bad to have the button show whatever album the current song is from as its another way to get more songs from the same context.

Kethku:
Another option could be to emulate spotify's experience. As far as I can tell the art in the spotify player shows the album but when clicked goes to whatever context the current song comes from (aka the play list, album radio, album, etc). IMHO this is confusing, but might make sense to just be what people expect coming from spotify

brandly:
@jpochyla I guess I've used Spotify so much that I didn't second guess the behavior at all! I definitely like seeing the extra color in that corner.

Looking more closely at Spotify's behavior, like @Kethku described, they generally show the current track's cover art regardless of where it's played from. The track title and album art are two separate links to the source it was played from, and the artist name links to the artist.

> I'm not sure it's always available, i.e. for playlist tracks.

It seems to not show up when playing directly from an album ironically, and I hoped the data plumbing could be wired up to get the image in place. Otherwise it already appears to be available even for playlists. 

> If we do it, it should probably act as a separate button, leading to the album view.

It's your call! If this project is thought of as _Native Spotify_, then maybe we mimic their UX. I think another option is breaking up the links a little differently -- album art to album, artist name to artist, and track name to playback source. 

Granted I don't think the current behavior is bad. You're viewing the _now playing_ content in that corner, and the entire link takes you back to its source. It's definitely more simple than Spotify's behavior, but when the mouse hovers over the cover art, the entire button lights up making it pretty clear it's a unified button.

Thanks for the feedback.

jpochyla:
Fair enough, thanks a lot for the extensive rationale! I've pushed a fix for the album tracks case, let's merge this! ❤️

PR #180, opened 2021-09-14
Fix compiling on nightly by bumping num-bigint

paolobarbolini:
More details: https://github.com/rust-num/num-bigint/issues/218


jpochyla:
Thanks!

PR #198, opened 2021-10-14
Playlist Manipulation: Adding tracks to playlists

martingoe:
As described in the commits, this PR is not quite ready to merge with the main project. 

I believe the following problems should be discussed:
Commit 4294be7:
- The current command executor for adding a track to a playlist is in the playlists `list_widget` function. This might not be the optimal position.
- If the request failed, the user should be notified. I believe this could be done with an alert box, but I did not currently find a way to implement this.

Commit 8efb405:
- The UI should be able to have access to the current user's ID without an API call.

Commit a43611e uses `Vec`s instead of the more commonly used `Vector`. 

edit: I was unaware of the draft feature github has but I believe this works just as well.

### DIFF
--- a/psst-core/src/item_id.rs
+++ b/psst-core/src/item_id.rs
@@ -1,4 +1,5 @@
 use std::{convert::TryInto, fmt, ops::Deref};
+use crate::error::Error;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum ItemIdType {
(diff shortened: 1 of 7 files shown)


jacksongoode:
I agree with @martingoe that an alert box would be essential - not just for this PR but for more transparency in general with failed API calls. I imagine it would be a bar that appears above the seek bar.

Does this PR include a method of removing tracks as well?

jpochyla:
Hi @martingoe, thanks a lot for your interest and for the PR! I finally have some more time so we can finish this. I'll push some more commits into the branch of this PR if that's fine with you.

martingoe:
Of course, go ahead. 
Unfortunately, I do have some busy weeks ahead myself so things might take a bit longer.

jpochyla:
I've added support for very simple error alerts in 731dadd.

jpochyla:
OK, so I've made some changes in a separate branch and merged to master, so far without the error alerts. Thanks a lot @martingoe and @jacksongoode.

PR #207, opened 2021-11-06
Bump most dependencies

paolobarbolini:
`cargo update`s all dependencies excluding `souvlaki` which in the [latest version](https://github.com/Sinono3/souvlaki/commit/5644d63af4e5c8c0ac9400fd0233cf5fc83096c7) adds a [stupid amount of useless dependencies](https://user-images.githubusercontent.com/6215781/140600040-d6782508-6711-4054-912b-2915dff73cd7.png)

PR #208, opened 2021-11-06
Replace `chrono` with the `time` crate

paolobarbolini:
Replaces the `chrono` crate with `time` 0.3

Depends on  #207


jpochyla:
Thanks, nice! #207 is merged, mind removing the commit from here? 

BTW if this PR would be based on top of the #207 branch, GitHub would automatically replant it on top of master, after merging.

paolobarbolini:
> Thanks, nice! #207 is merged, mind removing the commit from here?

Done

> BTW if this PR would be based on top of the #207 branch, GitHub would automatically replant it on top of master, after merging.

Oh cool, I never looked into how that worked. Thanks!

PR #218, opened 2021-11-18
Crank souvlaki to 0.4.4 fixing OpenBSD support

klemensn:


jpochyla:
Oops I'm really sorry, of course I forgot to remove the souvlaki pin from 53c97094d6f2bc6bc573460a20ac7529fced9907 and broke this. We've kept souvlaki at 0.4.0 because it drags a lot of needless dependencies with it, thanks to `zbus` :/ I'll figure something out, but bundling an async runtime because of a dbus library for media key support sounds terrible.

Sinono3:
Just want to make a quick update on this. I didn't realize how much dependencies zbus would bring in, the library's proc macros were very attractive and I didn't think about the consequences :star_struck (I also messed up in not changing the minor version, causing some breakage here and there)

But. good news! :) 

Souvlaki reverted to dbus-crossroads, dropping the zbus dependency and the async usage. Just compiled psst with the new version and it works correctly for me in Linux (no major changes have been made in Windows and macOS)

jpochyla:
Wow, thanks a lot @Sinono3, that's great! ❤️

PR #228, opened 2021-12-06
Fix windows build

zznixt07:
The `build.rs` file [here](https://github.com/jpochyla/psst/blob/cd55dfc8c38e3bc8c8ef725bfca48343ff000b26/psst-gui/build.rs#L5) still requires the logo.ico file.


jpochyla:
Thanks for the fix, I missed this :( The PNGs already contain the new icon, we should generate the ICO from them instead of using the old one. I'll have a chance to do it on Wednesday.

PR #229, opened 2021-12-09
chore: Build Windows ICO using ImageMagick

jstarpl:
This PR adds a step to use ImageMagick to build the Windows ICO file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,10 @@ jobs:
       if: ${{ runner.os == 'Windows' }}
       run: |
         choco install -y imagemagick.app --no-progress
-        Get-ChildItem -Path "${env:ProgramFiles}" | Where-Object {($_.Name -Like 'ImageMagick*')} | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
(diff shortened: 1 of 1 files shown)


jpochyla:
Uuu, thanks a lot @jstarpl! Unfortunately it seems the Windows build is failing on the ICO step (with `magick` not found).

jstarpl:
I'm gonna look into it some more and try to figure out why it can't find ImageMagick.

jstarpl:
I think the problem was that the "desktop" way of setting the PATH across steps is no longer supported. Supposedly, this is how it should be done: https://stackoverflow.com/questions/60169752/how-to-update-the-path-in-a-github-action-workflow-file-for-a-windows-latest-hos . I tried to apply that solution, let's see if it works now.

jstarpl:
Learning PowerShell is fun! `/sarcasm`. I think the correct string should now be written to the PATH GitHub variable.

jstarpl:
Learning by doing. Hopefully now it's correct.

jstarpl:
Okay, the PATH works now, I just missed a line creating a 32-bit 16x16 version.

PR #232, opened 2021-12-11
Added Non-Premium Spotify Account Support

Otus9051:
Took a while to figure out but here we are!

Features:
No Ads
No Skip Limit
No BS

Problems:
Need to set audio quality to normal on First run then restart `psst`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Fast Spotify client with native GUI, without Electron, built in Rust. Very early in development, lacking in features, stability, and general user experience. It is being tested only on Mac so far, but aims for full Windows and Linux support. Contributions welcome!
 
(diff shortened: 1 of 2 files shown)


jacksongoode:
I imagine this change could put this project at risk (TOS and such) - it might be best to maintain this as a fork?

jpochyla:
Hi, thanks for the contribution but I'm not interested in breaking the ToS like this, and don't want to play the cat&mouse game with Spotify.

roderickvd:
@jpochyla good call. I sent you an e-mail over at your Gmail address.

@Otus9051 please don't do this and remove your fork. While you might think this is pretty cool, things like this can threaten the entire open source Spotify community, not just psst, and so in the end bite you in the back too.

PR #233, opened 2021-12-11
Add track cover art in playlist and search results

iandwelker:
Screen Shot 2021-12-11 at 4 38 40 PM

This adds cover art images to tracks within playlists and search results (but not in albums), akin to the first party Spotify app (see attached image for a visual of what it looks like now). I didn't see an issue about this, but I feel it makes sense since that is what the spotify electron app does.

If you feel this is an unwanted change or should be implemented some other way, let me know and I'd be glad to discuss it :)

### DIFF
--- a/psst-gui/src/data/config.rs
+++ b/psst-gui/src/data/config.rs
@@ -75,6 +75,7 @@ pub struct Config {
     pub volume: f64,
     pub last_route: Option<Nav>,
     pub queue_behavior: QueueBehavior,
+    pub show_track_cover: bool,
 }
 
 impl Default for Config {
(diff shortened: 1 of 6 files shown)


confact:
I have been one of those who felt it was hard to read with the covers in a playlist. But Spotify has more information in the lists, so this could work. Otherwise, I would love to turn on/off this, as it is a bit distracting.

Great job, nonetheless!

jacksongoode:
Looks great, and if it could be toggleable for those who want more minimalism, that would be even better!

jpochyla:
Nice, thanks for the PR! Let's add a switch to the Preferences for this and merge?

jpochyla:
I realized it would be a bit more involved because of the `CommonCtx` stuff, so I did it myself. What do you think @iandwelker?

iandwelker:
> I realized it would be a bit more involved because of the `CommonCtx` stuff, so I did it myself. What do you think @iandwelker?

Looks good to me! I think that adding a toggle would've been a bit over my head, so I appreciate you doing so :)

PR #244, opened 2022-01-01
Navigate back when the 'back' button is pressed on mouse

gordonshieh:
QoL feature that I and maybe others will find useful when navigating.
Tested on my Logitech G203 on Windows. Pressing the "down" (aka back button; aka MouseButton::X1) will navigate to the previous page in the UI.


jpochyla:
Nice, thanks!

PR #247, opened 2022-01-05
Update souvlaki

klemensn:
Upstream tracked back to dbus-crossroads in order to avoid pulling
dbus as dependency into users like psst.

Done with `cargo update -p souvlaki` after changing psst-gui/Cargo.toml

Prompted by https://github.com/jpochyla/psst/pull/218#issuecomment-1003643688


jpochyla:
Thank you!

PR #260, opened 2022-01-16
Update Fedora dependencies in README.md

nilgradisnik:
CPAL library requires `alsa-lib-devel` package to be installed on Fedora

https://github.com/RustAudio/cpal#cpal---cross-platform-audio-library


jpochyla:
Thank you!

PR #261, opened 2022-01-16
Stop playback on stop command

forslund:
Currently the audio playback continues when the stop command is received in the player. To me this was quite counter-intuitive, and I think the playback should stop directly.

This adds a call to `self.audio_output_sink.stop()` when the stop-command is received.

This can easily be tested using the cli-client where "s" will not do anything.


forslund:
Read the code some more...perhaps it's better to run a `self.audio_output_sink.stop()` instead...

jpochyla:
Thanks!

PR #266, opened 2022-01-19
feat: implement scrollable volume slider

charliethomson:
From Spotify & YouTube, I missed scrolling to change the volume, so I implemented it. I added one field to the config 
image
with defaults 
image
image
You can also invert the scroll direction by making the x or y field negative


jpochyla:
Hi, thanks a lot! I'll be back to keyboard after a month of travels next week, so I'll jump on this asap.

Nojipiz:
What happened with this feature? :(

Issue #265, opened 2022-01-19
Artist Page: Move list of singles to immediately after list of albums


dmoa:
![example_spotify](https://user-images.githubusercontent.com/44374106/150220104-86b5f783-cb34-4790-836b-ef10231ffd69.png)

I'm aware singles are below the "appears on" list, but it may be more convenient to place singles above a less commonly used list.

charliethomson:
Looks like it's a Spotify thing. 
![image](https://user-images.githubusercontent.com/38197300/150235549-b4d6a925-954d-46a9-af74-5734fc333893.png)
They seem to be marking 'appears on' albums as 'albums' 

by no means is this a good way to solve this issue, but it works for coldplay as far as I can tell
![image](https://user-images.githubusercontent.com/38197300/150236061-44d6ba40-5138-415b-bc12-0e0c62f75d0f.png)
![image](https://user-images.githubusercontent.com/38197300/150235990-ec11c352-f551-4b32-9836-70debf1799a8.png)

I also added a section at the bottom for 'appears on' albums 
![image](https://user-images.githubusercontent.com/38197300/150236174-5f86e181-0c66-4405-8b4e-b25b62eda558.png)
![image](https://user-images.githubusercontent.com/38197300/150236141-3f609ac8-570f-4813-8fb3-2ddf1a8bc08d.png)

@jpochyla:
this is a bit of a hacky way to do this, any ideas?

jacksongoode:
Nice - I was just about to open an issue on this!

Insprill:
Closed with #271

PR #271, opened 2022-01-23
fix: albums that should be categorized as appears_on being categorized as album

charliethomson:
#265 

Open to feedback/changes. This was the best solution I saw in the ~2 hours of research I did


jacksongoode:
@charliethomson Does this still break in the case you mentioned in the comment? Did you experience any fail cases?

PR #282, opened 2022-02-25
very basic ico creation out of github actions

### LINKED ISSUE #259: Build fail: "Could not compile resource file" - Windows 10
**Describe the bug**
The application failed to build on my Windows 10 machine with the following error:
```
error: failed to run custom build command for `psst-gui v0.1.0 (C:\Rust\psst\psst-gui)`

Caused by:
  package.metadata.winres does not exist

  --- stderr
  gcc: error: C:Rustpssttargetreleasebuildpsst-gui-7cf215d7ed9b0c72outresource.rc: No such file or directory
  gcc: fatal error: no input files
  compilation terminated.
  /usr/bin/windres: preprocessing failed.
  thread 'main' panicked at 'Could not attach exe icon: Custom { kind: Other, error: "Could not compile resource file" }', psst-gui\build.rs:6:23
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**To Reproduce**
Ran the following commands in the `psst-gui` directory:
```
git submodule update --recursive --init
cargo build --release
```

**Expected behavior**
Build success.

**Screenshots**
N/A

**Environment**
 - OS: Windows 10 Education
 - Version: 2004

**Additional context**
My default cargo toolchain is stable-gnu but I also tried stable-msvc and nightly-gnu.

jacomago:
To fix #259 

Note I didn't add all of the other variations of resolution and color sizes, since from [ico format history](https://en.wikipedia.org/wiki/ICO_(file_format)) it seems that from Windows Vista resolution: 256x256 is supported with all colors. Also that is the point at which png are supported, so I don't think smaller icons are necessary.

### DIFF
--- a/psst-gui/Cargo.toml
+++ b/psst-gui/Cargo.toml
@@ -40,6 +40,7 @@ souvlaki = { version = "0.5.1" }
 
 [target.'cfg(windows)'.build-dependencies]
 winres = { version = "0.1.12" }
+ico = "0.1"
 
 [package.metadata.bundle]
 name = "Psst"
--- a/psst-gui/Cargo.toml
+++ b/psst-gui/Cargo.toml
@@ -40,7 +40,7 @@ souvlaki = { version = "0.5.1" }
 
 [target.'cfg(windows)'.build-dependencies]
 winres = { version = "0.1.12" }
-ico = "0.1"
+image = { version = "0.24"}
 
 [package.metadata.bundle]
 name = "Psst"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,26 +27,6 @@ jobs:
         profile: minimal
         toolchain: stable
 
-    - name: Install ImageMagick on Windows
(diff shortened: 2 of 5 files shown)


psst-gui/Cargo.toml
```diff
@@ -37,6 +37,7 @@ url = "2.2"
 
 [target.'cfg(windows)'.build-dependencies]
 winres = "0.1.11"
+ico = "0.1"
```
paolobarbolini, line 40: FWIW, considering also that it's already in the dependency tree and depends on up-to-date dependencies, image 0.24 supports generating an .ico containing multiple variants (sizes):
https://docs.rs/image/latest/image/codecs/ico/struct.IcoEncoder.html#method.encode_images

jacomago: Updated to image-rs

jpochyla:
Hi! Thanks a lot for the PR, and sorry for the delay. Can you please rebase on top of master? Merging afterwards!

jacomago:
Done!

PR #293, opened 2022-04-13
Add support for 'Remove from playlist'

iandwelker:
This adds support for the 'Remove from playlist' option when right-clicking on a track within a playlist. Clicking this button will remove the track from the playlist, show a banner at the bottom of the window stating if it succeeded or failed, and reload the list of tracks so that it is fully updated (ideally it would just remove that specific track from the list currently displayed, but I wasn't able to figure out how to do that :).

I like nearly everything about how I set this up, besides how I decided to implement storing the data about the current playlist. I added another parameter to `Track` called `current_playlist`, which stores the `PlaylistLink` of the currently displayed playlist (if it is being displayed in a playlist, otherwise `None`). This is not ideal for two reasons:
1. The `PlaylistLink` is stored once for every track in the playlist, which is just unnecessary data duplication since it is the same for every track.
2. The currently displayed playlist is UI data, not intrinsic data about the track itself.

Ideally, I would've liked to store the data about the current playlist in `track::Display`, but I couldn't find any way to pass that data in - it seems that `track::Display` is constructed without knowledge of any external context. Since the only other way to pass in data to the track's context menu is through the `Track` or `Library`, I thought that storing the data on `Track` would make the most sense. If it was stored on `Library`, there's a higher chance of it getting out-of-sync with what playlist is actually being displayed, and `Track` is destroyed when the view moves away, so we don't have to worry as much about the data not being tracked correctly.

If someone has better ideas about how to pass the data through, I welcome all suggestions :)

### DIFF
--- a/psst-gui/src/data/mod.rs
+++ b/psst-gui/src/data/mod.rs
@@ -42,7 +42,7 @@ pub use crate::data::{
         NowPlaying, Playable, PlayableMatcher, Playback, PlaybackOrigin, PlaybackPayload,
         PlaybackState, QueueBehavior, QueueEntry,
     },
-    playlist::{Playlist, PlaylistAddTrack, PlaylistDetail, PlaylistLink, PlaylistTracks},
+    playlist::{Playlist, PlaylistAddTrack, PlaylistRemoveTrack, PlaylistDetail, PlaylistLink, PlaylistTracks},
     promise::{Promise, PromiseState},
     recommend::{
         Range, Recommend, Recommendations, RecommendationsKnobs, RecommendationsParams,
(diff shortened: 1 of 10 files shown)


jacksongoode:
Does this work for collaborative playlists as well (potentially not owned by the user)?

jpochyla:
Hi @iandwelker, sorry for the hiatus and thanks a lot for the PR. I have an idea how to avoid the `current_playlist` back-link, I'll update the PR and merge.

jpochyla:
Ooops, missed the comment. I've pushed an attempt at removing the `current_playlist` link, but now I'm not sure about it all. This data smuggling is crazy, hopefully things will get better with the new Druid architecture.

iandwelker:
Ok, it should now work with collaborative playlists and should only show the 'Remove from playlist' option in the context menu if you're actually allowed to edit the playlist (before fe03126, it would show it on every playlist). I don't know if my implementation is exactly how it should be implemented (especially with the `Promise` type within the library - is `.resolved()` the method we should use to get the wrapped value?) and I feel good about merging it (once I fix formatting for CI, of course).

jpochyla:
Thanks a lot @iandwelker, and sorry for the wait 🤕

PR #297, opened 2022-04-30
Create makefile to make MacOS .app bundle

JoseVarelaP:
Given right now the applications builds as a unix executable, I wanted to make it compile and show itself as a proper .app bundle, this PR aims to do such thing.

Stuff to do:
- [X] Create icns and base .app bundle to generate the unix executable.
- [X] Create makefile that deals with copying the unix to the .app bundle.
- [ ] Make instance to perform the same action on a debug build
- [ ] Look into making the makefile not Mac-exclusive (if possible).

In action:
Captura de Pantalla 1


JoseVarelaP:
Nevermind this, I just noticed the build-actions already generates this via `cargo-bundle`.

PR #302, opened 2022-05-26
feat: set config file permissions to 0o600 on unix

cyyynthia:
This ensures credentials/auth tokens cannot be retrieved by other users.

I have little experience with Rust, especially with taking cross-platform into account, so I hope I did it right 😅


jpochyla:
Thank you!

PR #314, opened 2022-07-12
Added macOS brew installation

Just-Moh-it:
The title says it all. Since there's an existing brew cask for psst [here](https://formulae.brew.sh/cask/psst), it would be awesome to include install instructions in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,15 @@ RHEL/Fedora:
 sudo dnf install openssl-devel gtk3-devel cairo-devel alsa-lib-devel
 ```
 
+##### MacOS
+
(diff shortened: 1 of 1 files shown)


shustinm:
If you download the cask from brew and run it, you get the following error:
> “Psst” cannot be opened because the developer cannot be verified. macOS cannot verify that this app is free from malware.
So I had to open the Application folder, then right click Psst, and then click Open. Is it better to wait until it's verified before adding it to the README?

jacksongoode:
Not sure if that behavior is standard with unsigned apps, but if it is we can add the instructions?

PR #327, opened 2022-08-15
Fixes panic when using with Cubeb on Windows

SuisChan:
When running with Cubeb on Windows I get an error `fatal error: hr != CO_E_NOTINITIALIZED` 

This fixes the problem by initializing COM manually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2268,6 +2268,7 @@ dependencies = [
  "tempfile",
  "ureq",
  "url",
+ "windows 0.39.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jpochyla:
Thanks @SuisChan! Do you think we should open an issue at `cubeb`?

SuisChan:
> we should open an issue at `cubeb`?

Idk, but it looks like it's the user's responsibility. 

https://github.com/mozilla/cubeb-rs/issues/45#issuecomment-526959266

jpochyla:
Alright!

PR #330, opened 2022-08-23
Chunk fetching no longer requires 4-byte alignment

luukvanderduim:
As learned in discussion #308, Psst no longer require 4-byte alignment on fetching data chunks.

This removes functions `round_down_to_multiple` and `round_up_to_multiple`,
 and adjusts the `Read` implementation for `StreamReader` accordingly.


jpochyla:
Thank you!

PR #331, opened 2022-08-24
Fix conditional compilation of windows-specific code

mtorromeo:
The code should be compiled out, otherwise the build will only work on windows as the windows crate becomes a hard dependency


jpochyla:
Thanks, missed that one <3

PR #332, opened 2022-08-26
feat: add support for local file playback

### LINKED ISSUE #326: Playback of local files
We have _some_ local track support since #136, it'd be cool if we could play them.

superhawk610:
This PR adds support for local file playback. Local files have been visible in the UI, but not playable, since #136.

Resolves #326.

## Implementation

The primary struct used to identify a track for playback is `MediaPath`, which is composed of

- `item_id` a unique, opaque identifier, meaningful only to Spotify's API
- `file_id` a unique identifier that points into the local filesystem cache
- `file_format` an enum indicating what codec this file should be played with
- `duration` a `std::time::Duration` indicating the track's length

For remote files hosted on CDN, Spotify's API provides `item_id`, `file_format`, and `duration`, and the app stores `file_id` if/when it begins streaming the track. For local files, we don't need `file_id`, since the track already resides on disk, so it's left blank (`Default::default()`). We're able to guess `file_format` and `duration` by using Symphonia's `Probe`. Local files don't have an `item_id` on Spotify's end, so we have to generate our own. To do this, this PR includes `LocalItemRegistry`, which is a singleton that registers `PathBuf`s corresponding to local tracks and associates each with a unique ID that may be used in place of Spotify's item ID.

I can confirm that this works locally, though actually getting Symphonia to playback all tracks is a bit hit or miss (using an album from Bandcamp, about half the tracks work flawlessly and the other half won't play).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,9 +27,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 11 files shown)


jpochyla:
Thanks a lot, and sorry for the long wait :( I've changed some nits.

Issue #336, opened 2022-08-30
Use a logarithmic scale for volume


FichteFoll:
**Is your feature request related to a problem? Please describe.**
Trying to fit the volume of your music to, for example, another audio source that is playing in the back- or foreground is not an easy feat when you effectively only have a percent range of 0-15% to operate within as the rest hopelessly overshadows the other audio source.

**Describe the solution you'd like**
I suggest to use a logarithmic (e.g. dB-based) scaling of the volume slider. Spotify itself does this, as do other media players like foobar2k, mpv, and even YouTube recently adopted it.

**Describe alternatives you've considered**
It doesn't matter if the slider shows the dB or some arbitrary % (like spotify) because the latter will be easier to understand by most users and not raise questions.

jpochyla:
Hi, this got implemented in #339.

PR #339, opened 2022-09-27
Add logarithmic volume scaling

clin045:
Implement volume scaling (https://github.com/jpochyla/psst/issues/336). Instead of actually calculating the logarithm, I just used x^4 which is a good approximation suggested [here](https://www.dr-lex.be/info-stuff/volumecontrols.html).


jpochyla:
Thank you!

PR #341, opened 2022-10-08
Clear credentials from UI after authenticating

Insprill:
When logging in initially, or after changing accounts, until psst is restarted, when the settings panel is opened, the credentials will be visible in plain text. This clears the credentials from the UI after a login attempt, no matter if it's successful or not.

If #342 gets merged then perhaps this should only clear the credentials if authentication was successful?


jpochyla:
Hi, thanks a lot for all the PRs, great work!

> perhaps this should only clear the credentials if authentication was successful?

I agree.

Insprill:
What do you think about only clearing the password if authentication fails?

Issue #350, opened 2022-10-20
MacOS Native Media Player playback bar out-of-sync


literallyjustroy:
**Describe the bug**
On MacOS the native media player playback bar is not controllable and does not follow along with the same bar in the Psst application.

**To Reproduce**
Play any song via Psst on MacOS, then try to change to a different part of the song via the native MacOS media player.
As well, you can change the position in the song via Psst, and watch the progress not get reflected in the native media player.

**Expected behavior**
The native MacOS media player playback progress bar should allow you to control the position in the song.

**Screenshots**
<img width="496" alt="image" src="https://user-images.githubusercontent.com/56088145/196998421-620f04f7-8937-44c4-895e-c3e714e375d4.png">
<img width="293" alt="image" src="https://user-images.githubusercontent.com/56088145/196998459-725dffae-ee0b-4c33-a961-3eb772f13871.png">

**Environment**
 - OS: MacOS
 - Version: master branch 10/20/22, not sure where to find the version

**Additional context**
The [souvlaki](https://github.com/Sinono3/souvlaki/pull/28) `MediaControlEvent::SetPosition` event hasn't been implemented yet in Psst, as well, the MacOS implementation in souvlaki itself isn't completed. As well, the `update_media_control_metadata()` method is only being called during play/pause and needs to have some of its code pulled out and used during a `cmd::PLAY_SEEK` event.

literallyjustroy:
I'm working on this now. I've opened a [PR to implement the SetPosition event for MacOS in souvlaki](https://github.com/Sinono3/souvlaki/pull/28). Once that is handled I can open a PR with the rest of the changes listed in the **Additional Context**. Though depending on how long the souvlaki PR takes to get wrapped up I may work on the metadata updating as a separate branch.

Insprill:
Closed with #351 and #352.

PR #351, opened 2022-10-21
Sync native media playback progress bar on change from client

literallyjustroy:
This fixes the problem of the media player not being in-sync with changes from the Psst client. This resolves part of https://github.com/jpochyla/psst/issues/350. This does not fix the usability of the native media player progress bar, this will happen in a separate change (requires upstream change to dependency).

Before:
psst_media_player_sync_before

After:
psst_media_player_sync_after


jacksongoode:
Beautiful 🤩 !

PR #358, opened 2022-11-19
Fix queue behaviour when switching away from random

bivhitscar:
Playback should proceed according to the currently selected queue behaviour. The implementation of random/shuffle was causing other options to fail if it had been previously selected. 

This change restores the actual playlist position when switching queue behaviours, allowing for correct queue advancement at the end of the currently playing item.

To reproduce the original problem:
Open a playlist/album.
Select and play any item other than number 1*.
Select the random queue behaviour.
Select the loop all behaviour.
Click the skip to next item button.
When play advances to the next item, it will not be as expected (item 2 will probably play).

*The problem still exists from here, but is harder to spot.

### DIFF
--- a/psst-core/src/player/queue.rs
+++ b/psst-core/src/player/queue.rs
@@ -61,20 +61,17 @@ impl Queue {
         // Start with an ordered 1:1 mapping.
         self.positions = (0..self.items.len()).collect();
 
-        match self.behavior {
-            QueueBehavior::Random => {
-                // Swap the current position with the first item, so we will start from the
-                // beginning, with the full queue ahead of us.  Then shuffle the rest of the
-                // items and set the position to 0.
-                if self.positions.len() > 1 {
-                    self.positions.swap(0, self.position);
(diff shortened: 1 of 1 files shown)


psst-core/src/player/queue.rs
```diff
@@ -51,6 +51,13 @@ impl Queue {
     }
 
     fn compute_positions(&mut self) {
+        // In the case of switching away from shuffle, the position should be set back to
+        // where it appears in the actual playlist order.
+        let playlist_position = if self.positions.len() > 1 {
+            self.positions[self.position]
+        } else {
+            self.position
+        };
```
Insprill, line 60: `self.position` can be set to the result of this directly since this let isn't used anywhere else.

Insprill (changes requested):

psst-core/src/player/queue.rs
Insprill: Actually, this wouldn't work, as it depends on the randomized order of `positions`, and that gets reset on line 62.

PR #363, opened 2023-01-11
feat: view build info

cloudybyte:
Adding a way to check build info in the preferences

image


Insprill (changes requested):
Thank's for the PR! I think it'd be more fitting to add a new 'About' tab that could contain this, along with some other information like a short description of the project, a link to the repo, and the project's license.

Updating all the dependencies should also be done in its own PR.

cloudybyte:
I like the idea, but need an icon to go with it...
Are they an original creation for the project, or did you download them somewhere?

jacksongoode:
I agree, if this info could be moved into an info page, along with a link to the source page, that would be nice 🤔

cloudybyte:
Give me an icon and I should be able to finish this up in the next few days :)

Insprill:
I'm not sure where @jpochyla got the icons from, so I'd say just use a placeholder for now and we'll figure out the icon before merging.

arch-btw:
Hey guys, I'm new to this project. Let me start off by saying thank you so much! I use psst every day :-)

In regards to the icons used, it appears that it's `SF Symbols` by Apple.

You can download them here: https://developer.apple.com/sf-symbols/

The symbols used in the Psst menu are called `gearshape`, `key` and `internaldrive`

They are commented here as "SF Pro Regular", but I think that's supposed to be "SF Symbols" because the Pro Regular package doesn't have these symbols.

See here: https://github.com/jpochyla/psst/blob/master/psst-gui/src/widget/icons.rs

jacksongoode:
> Hey guys, I'm new to this project. Let me start off by saying thank you so much! I use psst every day :-)
> 
> In regards to the icons used, it appears that it's `SF Symbols` by Apple.
> 
> You can download them here: https://developer.apple.com/sf-symbols/
> 
> The symbols used in the Psst menu are called `gearshape`, `key` and `internaldrive`
> 
> They are commented here as "SF Pro Regular", but I think that's supposed to be "SF Symbols" because the Pro Regular package doesn't have these symbols.
> 
> See here: https://github.com/jpochyla/psst/blob/master/psst-gui/src/widget/icons.rs

Good detective work!

jpochyla:
Oops, thanks a lot, yup the icons are indeed from SF Symbols!

cloudybyte:
Sorry for the long delay, I've had some private stuff going on.
It seems that using Apples SF Symbols is against their license agreement, and thus not legal/at least against their ToS.

[This](https://stackoverflow.com/questions/64083993/is-there-a-way-to-use-the-apple-sf-symbols-on-the-web) StackOverflow post states that: 

> Your use of Symbols obtained from Apple’s SF Font is limited to creating mock-ups of user interfaces for software products running on Apple’s iOS, iPadOS, macOS and tvOS operating systems.

Although I wasn't able to verify/locate that sentence on an official Apple website.
I don't know whether they'll pursue us for packaging them, but personally tend to not risk it.

Apart from that, I don't have any way to access them anyways. I don't own a Mac, and they're only distributed as .dmg.

Any thoughts on how to move forward with this? @jpochyla @Insprill

cloudybyte:
This is what it looks like rn: 
image

The repo url is dynamically generated at compile time with the repos `origin` url. 
We obviously need another icon and probably change the tab's title.
As you can see the URL gets moved(/linewrapped?), but I can't figure out how to keep its position fixed.

Another idea I've had is copying the other two labels into the system's clipboard on click. Maybe a nice QoL feature for troubleshooting?

psst-core/build.rs
Insprill, line 1: Can we avoid creating a new file on disk for each build?

Insprill (changes requested):
We'll have to wait on @jpochyla for the icon issue.

psst-core/build.rs
cloudybyte: Its not creating a new file for every build, but overwriting the already existing one. We could include the info in one fie, but then we'd need to start parsing a format. I reckoned this to be the simplest solution.

The problem is, that there is no good way to pass info from the build context to the running application as of now.

PR #373, opened 2023-01-28
use buffered IO when loading and saving config

HookedBehemoth:
As per https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html

> The content of the IO stream is deserialized directly from the stream without being buffered in memory by serde_json.
> 
> When reading from a source against which short reads are not efficient, such as a [File](https://doc.rust-lang.org/std/fs/struct.File.html), you will want to apply your own buffering because serde_json will not buffer the input. See [std::io::BufReader](https://doc.rust-lang.org/std/io/struct.BufReader.html).

This is also true for `to_writer_pretty`. 

In my minimal testing, this shaved off 1000 syscalls and 15ms of the launch time. Especially reading was pretty bad as it read one byte at a time.

### DIFF
--- a/psst-gui/src/data/config.rs
+++ b/psst-gui/src/data/config.rs
@@ -1,3 +1,4 @@
+use std::io::{BufReader, BufWriter};
 use std::{env, env::VarError, fs::File, path::PathBuf};
 
 use std::fs::OpenOptions;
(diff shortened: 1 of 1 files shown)


Insprill (approved):
LGTM. Thank you!

PR #374, opened 2023-02-11
gui: Avoid unwrapping when fetching images

ptrcnull:
On bad connections, one timeout can cause the whole app to crash


ptrcnull:
thanks for the style fix :)

Insprill (approved):
LGTM. Thank you!

PR #379, opened 2023-03-17
Safer error-handling mechanism

arch-btw:
Hi, this is my first time writing Rust, please go easy on me 🙂 Could you make sure I did it correct? I won't be offended if it's not any good. This error was I think similar to the one found in https://github.com/jpochyla/psst/pull/374

The error was as follows and it would cause the app to crash and shutdown:

> thread 'main' panicked at 'called Result::unwrap() on an Err value: "SendError(..)"', psst-gui/src/controller/playback.rs:219:51

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -214,14 +214,14 @@ impl PlaybackController {
                 .unwrap();
         }
     }
-    
-fn send(&mut self, event: PlayerEvent) {
-    if let Some(s) = &self.sender {
-        s.send(event)
-            .map_err(|e| log::error!("Error sending message: {:?}", e))
-            .ok();
+
+    fn send(&mut self, event: PlayerEvent) {
+        if let Some(s) = &self.sender {
+            s.send(event)
+                .map_err(|e| log::error!("Error sending message: {:?}", e))
+                .ok();
+        }
     }
-}
 
     fn play(&mut self, items: &Vector<QueueEntry>, position: usize) {
         let items = items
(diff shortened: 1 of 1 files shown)


psst-gui/src/controller/playback.rs
```diff
@@ -215,9 +215,13 @@ impl PlaybackController {
         }
     }
 
-    fn send(&mut self, event: PlayerEvent) {
-        self.sender.as_mut().unwrap().send(event).unwrap();
+fn send(&mut self, event: PlayerEvent) {
+    if let Some(s) = &self.sender {
+        if let Err(e) = s.send(event) {
+            eprintln!("Error sending message: {:?}", e);
```
Insprill, line 221: Please use `log::error!` instead of `eprintln!`

Insprill (changes requested):
Thank you for the PR! Rust is fun :)

Please run `cargo fmt --all` before committing to ensure your code is styled properly, it doesn't look to be indented.

arch-btw:
Thank you so much for your help! I really appreciate it. I made the change and also did `cargo fmt --all`
Feel free to let me know if it's still not proper. 
Thanks again!

Insprill:
It looks like the indentation still isn't correct, not sure why fmt isn't fixing that. Please indent the method to match the others around it.

arch-btw:
Sorry about that, I think I messed it up but don't know why.
I ran `cargo fmt --all` again and it did change the file again, so I think it should work now. Thank you!

Insprill (approved):
LGTM. Thank you!

PR #384, opened 2023-04-12
README update: song highlighting implemented

cloudybyte:
I don't really know, what you meant by this, but tried it out on my local machine and it seems to be working fine.
Did someone just forget to update the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,9 @@ cargo run
 - [x] Audio loudness normalization
 - [x] Genre playlists and "For You" content
 - [x] Dark theme
+- [x] Correct playback highlight
+    - Highlight now-playing track only in the correct album / playlist
+    - Keep highlighted track in viewport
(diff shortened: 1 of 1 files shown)


Insprill:
@jpochyla Can you confirm what you mean here?

jpochyla:
Hi, sure!

> Highlight now-playing track only in the correct album / playlist

I don't really know how to approach this. Track is always playing in a context (album, top tracks, search etc), and I think that now-playing track displayed in the now-playing context should look different from track that is technically now-playing (has the same track ID), but is displayed in a different context.

> Keep highlighted track in viewport

More straightforward. Auto-scroll to reveal tracks when they start to play, if the mouse cursor is not active (hovering the window or maybe not moving for a while?).

Insprill:
Closing since those aren't implemented yet

Issue #382, opened 2023-04-04
[Windows] "Failed to fill whole buffer" when trying to sign in


StoppingBuck:
**Describe the bug**
When trying to log in, the app displays "Failed to fill whole buffer". Nothing else happens

**Screenshots**
![image](https://user-images.githubusercontent.com/8426548/229861066-e371f8fc-d9d4-45ac-a22f-f68216420669.png)

**Environment**
 - OS: Windows 10
 - Version: Latest master build as per 2023-04-04

StoppingBuck:
It is worth mentioning that Windows SmartScreen tried blocking the .exe from being run, and I had to manually select "Run Anyway"

cloudybyte:
Hiding the password/using a protected textbox is not possible as of now, as druid (the GUI framework we use) doesn't support it. There was some discussion around it (and even some first attempts at implementing,  see linebender/druid#1839, linebender/druid#1114 and linebender/druid#2105), but I don't think it's worth it to invest the time as druid is discontinued.

However, I've fixed the broken error message in #386

PR #386, opened 2023-04-13
fix: fix error message on login w/o creds

cloudybyte:
addressing #382

### DIFF
--- a/psst-core/src/connection/mod.rs
+++ b/psst-core/src/connection/mod.rs
@@ -234,9 +234,9 @@ impl Transport {
         // check for empty username & password:
         if credentials.username.is_empty() && credentials.auth_data.is_empty() {
             return Err(Error::AuthFailed {
-                // code 12 = bad redentials
-                code: 12
-            })
+                // code 12 = bad credentials
+                code: 12,
+            });
         }
 
         // Send a login request with the client credentials.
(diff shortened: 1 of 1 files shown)


psst-core/src/connection/mod.rs
Insprill, line 1: Should this instead be checked [when the user clicks the log in button](https://github.com/jpochyla/psst/blob/master/psst-gui/src/ui/preferences.rs#L318)?

Insprill (changes requested):

psst-core/src/connection/mod.rs
cloudybyte: I don't think so. Checking for error conditions at the lowest level possible is good practice.
Otherwise reusing the fn in another context would result in the bug occurring again.

cloudybyte: What do you think? @Insprill @jpochyla

Insprill: It could only happen if someone manually modifies the config after logging in, and replaces username or auth_data with an empty string which seems... unlikely, but I suppose it can be handled here.

As an extension to that theoretical scenario, if that's the case, it should probably open the login screen instead of having errors all over the UI.

psst-core/src/connection/mod.rs
```diff
@@ -231,6 +231,14 @@ impl Transport {
     pub fn authenticate(&mut self, credentials: Credentials) -> Result<Credentials, Error> {
         use crate::protocol::{authentication::APWelcome, keyexchange::APLoginFailed};
 
+        // check for empty username & password:
+        if credentials.username.is_empty() && credentials.auth_data.is_empty() {
```
Insprill, line 235: Shouldn't this be a `||` instead of a `&&`?

Insprill (changes requested):

PR #387, opened 2023-04-13
Fixing index out of bounds error

arch-btw:
Hey :+1: 

Fixed an "index out of bounds" error in the play function of playback.rs by checking that the provided position value is within bounds before sending the LoadQueue command to the player. This was the error that I kept getting, causing the app to become non-responsive until restarting it:
```
thread '<unnamed>' panicked at 'index out of bounds: the len is 2 but the index is 10', 
psst-core/src/player/queue.rs:57:13 
```
I tried modifying some things in queue.rs first but that was the wrong approach.


Insprill (approved):
LGTM, though we should figure out why this can happen in the first.

Issue #347, opened 2022-10-11
Crossfade on play/stop


Geobert:
It's a small detail, but when using Psst, I felt that the play/stop was a bit aggresive and after a while, I nailed what was bothering me: the lack of crossfading on play/stop so that's my feature request :)

jacksongoode:
Would be a nice setting to toggle as well!

PR #390, opened 2023-04-21
Crossfade UI Implementation

podojilr:
Attempted to address #347. Unable to implement crossfading itself, but added Crossfade setting to preferences for future implementations of crossfading. Done in collaboration with @Javeh


Insprill:
I appreciate the PR, however, just implementing the UI side of things doesn't do a whole lot other than possibly be confusing.

Insprill:
I'm going to close this for now, however, if crossfading itself gets implemented, this would be a great headstart for the UI.

PR #392, opened 2023-04-21
Playlist Sorting

### LINKED ISSUE #372: Playlist sorting
I have large playlists which I always sort by recently added on mobile. Currently to play anything on them I've got to scroll all the way to find my recent things, and I have to do that each time the playlist reloads.

It would be real nice if there was a switch at the top of the playlist to sort it in this fashion.

It would be best if one could sort by all the things that spotify offers (author/song/album(?) name, date added, etc..) and see these properties visually, but just to sort by date added doesn't necessarily have to involve a ton of UI work.

image

Javeh:
Resolves #372 

I implemented the feature as described, adding an arrow that indicates descending or ascending order. You can also right click the arrow to select different sorting criteria. It defaults to date added and ascending order, which is the default order of both Spotify and Psst. I implemented the sorting criteria in Spotify, plus sorting by artist name. Sort preferences persist across playlists and across boots of the application.

image
image
image

A potential improvement would be to manipulate the underlying data structure of a playlist to sort the tracks, rather than just refreshing the playlist and sorting the tracks at the time they are fetched, which is what this does.

Implemented in collaboration with @podojilr


Insprill (approved):
Thank you for the PR! Sorting the playlist locally would be better, but this works fine for now. LGTM!

PR #398, opened 2023-04-30
Compress images with TinyPNG

Insprill:
Compresses the logo and application screenshot using [TinyPNG](https://tinypng.com), saving about 512KB in total.


jacksongoode (approved):
I assume this is lossless compression? If so, seems good.

JulienMaille:
> I assume this is lossless compression? If so, seems good.

TinyPNG uses smart lossy compression techniques to reduce the file size of your WEBP, JPEG and PNG files. By selectively decreasing the number of colors in the image, fewer bytes are required to store the data. The effect is nearly invisible but it makes a very large difference in file size!

PR #399, opened 2023-05-02
Add follow and unfollow playlist support

takanoro:
Users can now follow and unfollow playlists. Updated the UI [context menu] to include buttons for following and unfollowing playlists.

### DIFF
--- a/psst-gui/src/ui/playlist.rs
+++ b/psst-gui/src/ui/playlist.rs
@@ -362,7 +362,7 @@ fn playlist_menu_ctx(playlist: &WithCtx<Playlist>) -> Menu<AppState> {
         .command(cmd::COPY.with(playlist.url())),
     );
 
-    if library.contains_playlist(playlist) {
+    if library.is_created_by_user(playlist) {
         if library.is_owner(playlist) {
             let unfollow_msg = UnfollowPlaylist {
                 link: playlist.link(),
--- a/psst-gui/src/ui/playlist.rs
+++ b/psst-gui/src/ui/playlist.rs
@@ -7,6 +7,7 @@ use druid::{
 };
 use itertools::Itertools;
 
+use crate::data::WithCtx;
 use crate::{
     cmd,
     data::{
@@ -27,6 +28,9 @@ pub const LOAD_DETAIL: Selector<(PlaylistLink, AppState)> =
 pub const ADD_TRACK: Selector<PlaylistAddTrack> = Selector::new("app.playlist.add-track");
 pub const REMOVE_TRACK: Selector<PlaylistRemoveTrack> = Selector::new("app.playlist.remove-track");
 
+pub const FOLLOW_PLAYLIST: Selector<Playlist> = Selector::new("app.playlist.follow");
+pub const UNFOLLOW_PLAYLIST: Selector<PlaylistLink> = Selector::new("app.playlist.unfollow");
+
 pub fn list_widget() -> impl Widget<AppState> {
     Async::new(
         utils::spinner_widget,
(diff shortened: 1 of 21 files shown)


Insprill (changes requested):
Thank you for the PR! In Spotify, there's a distinction between playlists you've created and those you've followed. You can't unfollow a playlist you created, only delete it. With this PR, if you unfollow a playlist you created, it'll still exist, but won't show in your library. If I'm misunderstanding something, please correct me, but I don't think this is intended behavior.

takanoro:
Yes, you are correct about the behavior related to unfollowing a playlist you've created. So far, I haven't seen an option to delete such a playlist.

On [spotify web api concepts](https://developer.spotify.com/documentation/web-api/concepts/playlists) page spotify says:

> We have no endpoint for deleting a playlist in the Web API; the notion of deleting a playlist is not relevant within the Spotify’s playlist system. Even if you are the playlist’s owner and you choose to manually remove it from your own list of playlists, you are simply unfollowing it. Although this behavior may sound strange, it means that other users who are already following the playlist can keep enjoying it. [Manually restoring a deleted playlist](https://www.spotify.com/se/account/recover-playlists/) through the Spotify Accounts Service is the same thing as following one of your own playlists that you have previously unfollowed.

As a workaround we can omit "Unfollow" menu item for playlists you've created while retaining the ability to follow and unfollow other playlists.

Insprill:
Looking at it again, I think the current behavior is fine, and I was just misunderstanding Spotify's UI. As it says in the documentation you referenced, there's no such thing as deleting a playlist. When the Spotify UI says 'Delete', it simply unfollows it, and it can still be accessed and refollowed again if you know its ID.

The only other critique I have is the fact that there's no confirmation before deleting a playlist. I can easily see someone deleting a playlist they meant to copy the link to, and having no way of finding it again. This could be implemented by creating a small child window, similar to how the preferences window is handled, that just has some text and Yes/ No buttons. Implementing that will be a little complex, so if anyone has another idea feel free to share.

takanoro:
I've added a confirmation dialog window, but since I'm not very familiar with Druid, I may have made some kludges  :police_car: concerning the handling of events in inappropriate places (new window creation and their environment). Additionally, I'm uncertain about the padding and window size of the confirmation dialog window. 

Please, feel free to provide any corrections or suggestions, as I appreciate your input.

psst-gui/src/ui/playlist.rs
```diff
@@ -104,19 +150,87 @@ pub fn list_widget() -> impl Widget<AppState> {
     )
 }
 
-pub fn playlist_widget() -> impl Widget<Playlist> {
-    let playlist_image = rounded_cover_widget(theme::grid(6.0));
+fn confirmation_window(msg: UnfollowPlaylist) -> WindowDesc<AppState> {
+    let win = WindowDesc::new(confirmation_widget(msg))
+        .window_size((theme::grid(45.), theme::grid(20.)))
+        .title("Rename playlist")
```
Insprill, line 156: I don't think this is correct

psst-gui/src/ui/playlist.rs
```diff
@@ -104,19 +150,87 @@ pub fn list_widget() -> impl Widget<AppState> {
     )
 }
 
-pub fn playlist_widget() -> impl Widget<Playlist> {
-    let playlist_image = rounded_cover_widget(theme::grid(6.0));
+fn confirmation_window(msg: UnfollowPlaylist) -> WindowDesc<AppState> {
+    let win = WindowDesc::new(confirmation_widget(msg))
+        .window_size((theme::grid(45.), theme::grid(20.)))
```
Insprill, line 155: The height is a little too small to see the buttons. 25 seems to be good.

psst-gui/src/ui/playlist.rs
```diff
@@ -104,19 +150,87 @@ pub fn list_widget() -> impl Widget<AppState> {
...
+            ctx.window().close();
+        });
+    let cancel_button = Button::new("Cancel")
+        .fix_height(theme::grid(5.0))
+        .padding(theme::grid(1.5))
+        .on_click(|ctx, _, _| ctx.window().close());
+
+    let button_row = Flex::row()
+        .with_child(cancel_button)
+        .with_child(delete_button)
```
Insprill, line 210: The delete button should go first, seems more intuitive.

psst-gui/src/ui/playlist.rs
```diff
@@ -245,5 +362,46 @@ fn playlist_menu(playlist: &Playlist) -> Menu<AppState> {
         .command(cmd::COPY.with(playlist.url())),
     );
 
+    if library.contains_playlist(playlist) {
+        if library.is_created_by_user(playlist) {
+            let unfollow_msg = UnfollowPlaylist {
+                link: playlist.link(),
+                created_by_user: true,
+            };
```
Insprill, line 370: This can be moved up with `created_by_user` being set to `library.is_created_by_user(playlist)`

Insprill (changes requested):

takanoro:
Thank you for the review! 
It seems that I have incorporated all the suggestions into the code and corrected the typos.

Insprill (approved):
LGTM. Thank you!

PR #403, opened 2023-05-04
Add save button in track row and playback bar

### LINKED ISSUE #248: Star / Favorite
Really good work so far! I actually had to switch from Spotify Desktop because there's a strange network configuration problem I'm having with Linux, but it's only a real showstopper for just that one app and no others on my system, so I figured I'd put off figuring that out if I could just use Psst instead.

That said, I really miss the feature to star / favorite certain tracks I've been listening to.

- Star currently playing track, maybe to to the left of the rewind button on the player at the bottom section of the screen
- Star playlist track, on the far right of the screen, adjacent to the track length
- Each star should indicate the current favorite status
- It should also be able to toggle between both states

I could also just use the web player at open.spotify.com, but it's a bit of a pain to load on an older laptop just for the one thing.

**Additional context**
Maybe something like this (forgive my meager inkscape skills):
path1246
Maybe a star should appear for every track now that I think about it, considering it might be bothersome to track hover states.

Insprill:
Closes #248

image


jacksongoode:
Hmm, just from a UI perspective maybe the heart could be a touch smaller? Like maybe 20pt rather than 24?

Insprill:
Agreed. I made them a little smaller and updated the screenshot.

PR #408, opened 2023-05-05
Update dependencies

### LINKED ISSUE #153: Title bar remains white in dark theme
image

This does not look very nice. It should ideally be dark in dark theme (or follow whatever theme OS is in). See the explorer at the right for example.

Insprill:
Updates serde to 1.0.162, and our fork of druid to the latest master which includes a fix for window titlebar's being opposite of the OS theme on Windows.

Closes #153.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,19 +631,19 @@ checksum = "bd0c93bb4b0c6d9b77f4435b0ae98c24d17f1c45b2ff844c6151a07256ca923b"
 
 [[package]]
 name = "displaydoc"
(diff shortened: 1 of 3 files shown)


jacksongoode:
@Insprill Maybe this is a topic for a discussion but since druid has stopped development is the move to just continue bugfix development on the fork?

Insprill:
They're still accepting bug fixes, so we might as well contribute them upstream. If they don't get accepted for whatever reason, or they fully stop development, then we'll just apply it to our fork.

PR #409, opened 2023-05-05
Implemented playlist rename functionality

takanoro:
Added rename playlist support + tried to standardize styles for rename and unfollow dialog windows.


Insprill (changes requested):
Looks good, however, instead of reloading playlists from the API when they're renamed, we can just update the local name of it. That can be implemented similarly to how saved tracks and albums are updated at https://github.com/jpochyla/psst/blob/master/psst-gui/src/data/mod.rs#L293-L305. This would fix the issue of the playlist name being one-rename out of date, most likely due to Spotify's API not updating in time for the load request.

takanoro:
Gotcha, seems like it should be good now

Insprill (approved):
LGTM. Thank you!

PR #411, opened 2023-05-10
Split playlist & control section

jacksongoode:
This adds proper spacing between the volume control and divider and splits the playlist section and volume/controls section to provide a divider.

Here an image of this PR (left) and before (right):

Screen Shot 2023-05-09 at 9 00 42 PM

This also:
- Reduces padding around gear
- Increases the minimum height of window so that playlists can be visible at min height

### DIFF
--- a/psst-gui/src/ui/mod.rs
+++ b/psst-gui/src/ui/mod.rs
@@ -42,7 +42,7 @@ pub mod utils;
 pub fn main_window(config: &Config) -> WindowDesc<AppState> {
     let win = WindowDesc::new(root_widget())
         .title(compute_main_window_title)
-        .with_min_size((theme::grid(65.0), theme::grid(46.0)))
+        .with_min_size((theme::grid(65.0), theme::grid(50.0)))
         .window_size(config.window_size)
         .show_title(false)
         .transparent_titlebar(true);
--- a/psst-gui/src/ui/mod.rs
+++ b/psst-gui/src/ui/mod.rs
@@ -42,7 +42,7 @@ pub mod utils;
 pub fn main_window(config: &Config) -> WindowDesc<AppState> {
     let win = WindowDesc::new(root_widget())
         .title(compute_main_window_title)
-        .with_min_size((theme::grid(65.0), theme::grid(50.0)))
+        .with_min_size((theme::grid(65.0), theme::grid(65.0)))
         .window_size(config.window_size)
         .show_title(false)
         .transparent_titlebar(true);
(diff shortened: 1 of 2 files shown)


Insprill (commented):
Thank's for the PR! The code looks good; however, I'm unsure of the Split usage. I have only worked with a few Splits in Druid, so correct me if I'm wrong, but the split point is always relative to window size. This causes it to become disproportionally large or small, depending on the window size. For example, here's the window, with this PR, at a small - but reasonable size
image
There's plenty of space in the playlists section, but the username and preferences get cut off.

Here's another example with the window fullscreen 
image
There is lots of extra padding on the top/bottom of the elements that eats up the playlists section and doesn't look right.

While I like this change, it looks much better at specific window sizes; it would be more of a regression unless we can give the Split a fixed split point.

jacksongoode:
Yeah, I'm also unfamiliar with it but border wouldn't be able to create that nice division. I think there's some way to just set the split to the fixed size of one of the children. This was just a quick attempt and I'll see if there's some way to anchor the `controls` size.

jacksongoode:
@Insprill I was a little confused as to how a fixed container and flex container live together in a split? What is the non-flex way to make a container?

jacksongoode:
I believe I've resolved the issues you've mentioned @Insprill. I switched to a fixed stack in a column and added a border to the top of the controls.

psst-gui/src/ui/mod.rs
```diff
@@ -42,7 +42,7 @@ pub mod utils;
 pub fn main_window(config: &Config) -> WindowDesc<AppState> {
     let win = WindowDesc::new(root_widget())
         .title(compute_main_window_title)
-        .with_min_size((theme::grid(65.0), theme::grid(25.0)))
+        .with_min_size((theme::grid(65.0), theme::grid(65.0)))
```
Insprill, line 45: I think a better minimum height would be `46.0`. It allows just enough room for one playlist to be displayed without anything else getting squished.

Insprill (changes requested):

psst-gui/src/ui/mod.rs
Insprill: If it was the default size I could see that, but it's the minimum, so I think it should be as small as possible while still being usable.

Personally, I only have one playlist, so it would end up with a bunch of extra space that isn't being used.

jacksongoode: Okay, I'm okay with this. In a future PR I think we ought to hide the psst icon after some height threshold to free up space.

jacksongoode: I don't think 46 makes a single playlist possible to view... I think 50 would work considering double-lined playlists as well.

Insprill (approved):
LGTM

PR #413, opened 2023-05-11
Fix the window getting smaller each time it's opened

Insprill:
When the window is created, there's an initial WindowSize event that's slightly smaller than the desired window size. This causes the app to keep getting smaller each time it's opened until it's hit the minimum window size. This makes it so the first WindowSize event doesn't save its size to the config.

### DIFF
--- a/psst-gui/src/delegate.rs
+++ b/psst-gui/src/delegate.rs
@@ -16,6 +16,7 @@ pub struct Delegate {
     main_window: Option<WindowId>,
     preferences_window: Option<WindowId>,
     image_pool: ThreadPool,
+    size_updated: bool,
 }
 
 impl Delegate {
@@ -26,6 +27,7 @@ impl Delegate {
             main_window: None,
             preferences_window: None,
             image_pool: ThreadPool::with_name("image_loading".into(), MAX_IMAGE_THREADS),
+            size_updated: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


jacksongoode:
Wait why is it smaller than the expected size? What's actually causing the window to be smaller?

jacksongoode:
Maybe this might have to do with the fact the window location is not being saved as well?

Insprill:
I don't think it's due to the location not being saved. When the application starts, the saved size is correct. However, immediately after, it gets resized a little smaller. I thought maybe it was related to one size including the title bar and one size not, but that doesn't explain the width shrinking. I tried digging around in Druid to figure out where the resize event is even coming from, but I didn't have much luck.

psst-gui/src/delegate.rs
```diff
@@ -150,7 +152,12 @@ impl AppDelegate<AppState> for Delegate {
     ) -> Option<Event> {
         if self.main_window == Some(window_id) {
             if let Event::WindowSize(size) = event {
-                data.config.window_size = size;
+                // This is a little hacky, but without it, the window will slowly get smaller each time the application is opened.
+                if !self.size_updated {
+                    self.size_updated = true;
+                } else {
+                    data.config.window_size = size;
+                }
```
jacksongoode, line 160: It would be great to keep an issue related to this open so we can figure out a proper solution.

jacksongoode (approved):

arch-btw:
@Insprill thanks for all the awesome commits lately!

PR #422, opened 2023-05-19
fix: clears playlist positions to zero before recalculating on changing playlist (fix #420)

### LINKED ISSUE #420: Queue.fill() crash when switching playlists
**Describe the bug**
When switching from a short playlist to a long playlist by starting a track at the end of the long playlist, Queue.fill/compute_positions crashes:
```
thread '<unnamed>' panicked at 'index out of bounds: the len is 24 but the index is 29', psst-core/src/player/queue.rs:57:13
```

**To Reproduce**
Start playing a relatively short playlist, then start the last track of a longer playlist.

**Expected behavior**
The new track starts playing.

**Screenshots**
N/A

**Environment**
 - OS: Linux
 - Version: current git master; f94af14aa9fdd3d59cd92849fa7f076103b37a70

**Additional context**
I think adding a line `self.positions.clear()` at the start of Queue.fill() (psst-core/src/player/queue.rs) should fix the issue. The problem seems to be that it sets self.position while compute_positions() is still assuming self.position points into the now-outdated self.positions vector.

cameronraw:
This is a tiny change to fix the bug mention in issue #420 whereby an out of bounds error appears due to playlist lengths not being returned to zero upon changing to a different playlist.

### DIFF
--- a/psst-core/src/player/queue.rs
+++ b/psst-core/src/player/queue.rs
@@ -40,6 +40,7 @@ impl Queue {
     }
 
     pub fn fill(&mut self, items: Vec<PlaybackItem>, position: usize) {
+        self.positions.clear();
         self.items = items;
         self.position = position;
         self.compute_positions();


Insprill (approved):
LGTM. Thank you for the fix!

arch-btw:
Oh wow, thank you @cameronraw! I was trying to fix this in https://github.com/jpochyla/psst/pull/387 but as @Insprill pointed out, I still had to figure out _how_ it was happening. Turns out, this is it haha.

cameronraw:
No problem, @arch-btw ! The change was already suggested in the original issue, it was a nice and easy.

PR #191, opened 2021-10-06
Support OpenBSD

klemensn:
Track merge requests in dependencies/crates and update documentation accordingly.

All patches required to run psst on OpenBSD/amd64 have been submitted upstream:
- https://github.com/ExPixel/miniaudio-rs/pull/75
- https://github.com/linebender/druid/pull/1993
- https://github.com/linebender/piet/pull/466

Dependencies have been carefully patched to merely integrate those patches
required for OpenBSD support, i.e. no new features or unrelated changes are
pulled in.

This makes `cargo build` produce a working `./target/debug/psst-gui`.

This is already enough to create an OpenBSD port (from my fork) and provide
working binary packages to users.

Input on how to best incorporate all this would be welcome, hence this draft.

PR #426, opened 2023-05-28
README: Add build instructions for OpenBSD

klemensn:
As of today's HEAD, this produces `psst-gui` that at least starts and
prompts for account details, on OpenBSD/amd64 7.3 -current with
gtk+3-3.24.38 cairo-1.17.8p0 llvm-13.0.0p5

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,32 @@ RHEL/Fedora:
 sudo dnf install openssl-devel gtk3-devel cairo-devel alsa-lib-devel
 ```
 
+##### BSD [WIP]
+
+Similar to Linux, Druid defaults to GTK while providing an X11 backend as well.
+Furthermore, bindgen must be able to find LLVM through the expected environment variable.
+Only OpenBSD/amd64 has been tested so far.
+
+OpenBSD:
+```shell
+doas pkg_add gtk+3 cairo llvm
+export LIBCLANG_PATH=/usr/local/lib
+```
+
+In case rustc(1) fails building bigger crates
+```shell
+memory allocation of 2880 bytes failed
+error: could not compile `gtk`
+
+Caused by:
+  process didn't exit successfully: `rustc --crate-name gtk [...]` (signal: 6, SIGABRT: process abort signal)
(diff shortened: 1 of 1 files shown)


klemensn:
#191 is still an ongoing effort, but this would've helped me to get started again when coming back to it after two years.

Insprill (changes requested):
Glad to hear you're still working on BSD support!

There should be a disclaimer that BSD support is still a WIP, and isn't ready to be used yet.

The code block related to building is also a little tricky to understand. It should be separated out into individual commands/results instead of grouping it all together.

klemensn:
> Glad to hear you're still working on BSD support!
> 
> There should be a disclaimer that BSD support is still a WIP, and isn't ready to be used yet.
> 
> The code block related to building is also a little tricky to understand. It should be separated out into individual commands/results instead of grouping it all together.

Is this any better?

README.md
```diff
@@ -35,6 +35,36 @@ RHEL/Fedora:
...
+warning: build failed, waiting for other jobs to finish...
+```
+try increasing your user's maximum heap size:
+```shell
+ulimit -a | grep data
+data(kbytes)       1572864
+```
+```shell
+ulimit -d  $(( 2 * 1572864 ))
+```
```
Insprill, line 66: Could this be made into a one-liner? Any adverse side-effects should also be noted,

Insprill (changes requested):
Sorry about the slow review. I just have one last nitpick

README.md
klemensn: > Could this be made into a one-liner? Any adverse side-effects should also be noted,

I folded them.

These [ulimits](https://man.openbsd.org/ksh.1#ulmit) are limited to the shell one sets them from, and its children, i.e. cargo(1) and rustc(1), of course.

I am not aware of bad side effects -- you just get to allocate twice as much heap data per process from this shell.
If processes reach that limit, they themselves get killed;  afterall, this is not Linux where OOM may kill any process.

Insprill (approved):
LGTM!

PR #429, opened 2023-06-01
Remove the rust-toolchain action

Insprill:
Rust is [pre-installed on GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#preinstalled-software), so there's no need for us to have an action to install it explicitly.


jacksongoode:
Good catch!

Issue #412, opened 2023-05-10
Overlapping album hover selection on min window width


jacksongoode:
**Describe the bug**
Overlapping album hover selection on min window width.

**To Reproduce**
Reduce window to min and hover

**Expected behavior**
The heart/save should be included within the hoverable component.

**Screenshots**
<img width="154" alt="Screen Shot 2023-05-10 at 12 31 53 PM" src="https://github.com/jpochyla/psst/assets/54308792/e124ba2f-57cc-40b6-85f8-70dc501bb15a">

**Environment**
 - OS: macOS
 - Version: latest nightly

Insprill:
This also occurs with the track row, but the button can't be part of the flex because the click event gets propagated to the widget under it. If anyone has an idea of how to solve that, PRs are welcome!

SeaSpr:
I think I figured out what's going on here. The playing_item_widget (the hoverable thing)'s width is going to be what's leftover from the playback panel's width once the player_widget's (the controls) width is accounted for. The player_widget's width is constant, so what the playing_item_widget gets is wide enough to fit the left edge of the heart icon (triggering the heart to be rendered) but not wide enough to fit the entire heart.

Here's a screenshot of the panel with borders around the widgets. The blue's width is constant, the red's width is (total width - blue width). The padding on the sides of the controls is included in blue's width, which further eats into red's width that could cover the heart icon. There's also a gap between the album art and the heart that could be reduced.
![image](https://github.com/jpochyla/psst/assets/8027640/ecbd4b8b-b73e-4699-ae5c-eff9e6acd329)

With that in mind, I can think of 3 ways to fix the issue if the heart should stay visible.

1. Reduce the spacing between the heart and the album art to fit the heart at the minimum width.
![image](https://github.com/jpochyla/psst/assets/8027640/2e9ca892-3846-4323-bab6-0954fbe8febe)

3. Reduce the padding of the controls so that the playing_item_widget has more room at the minimum window width
4. Increase the minimum width of the window

Here is a PR that reduces the width of the playing_item_widget by making the spacer between the album art and info collapsible. Option 3 is probably the most elegant solution (1 line) but perhaps that's more of a bandaid than a fix.

https://github.com/jpochyla/psst/pull/432

Insprill:
Closed by #432.

jacksongoode:
@SeaSpr Thank you!

PR #432, opened 2023-07-26
Move spacer between cover art and track info to flex row

SeaSpr:
Moving the spacer to a flex row along with the track info makes the spacer collapsible, which frees space for the heart icon.

Fix Overlapping album hover selection on min window width #412


Insprill (approved):
LGTM. Thank you for the fix!

PR #436, opened 2023-09-08
Increase timeout when using proxy

arch-btw:
On large playlists and/or high-latency networks there is not enough time for all album covers to finish downloading. 

> WARN  psst_gui::delegate] failed to fetch image: https://i.scdn.co/image/... Connection Failed: tls connection init failed: Resource temporarily unavailable (os error 11)

Thank you :smiley:

### DIFF
--- a/psst-core/src/util.rs
+++ b/psst-core/src/util.rs
@@ -5,9 +5,9 @@ use quick_protobuf::{BytesReader, MessageRead, MessageWrite, Writer};
 
 use crate::error::Error;
 
-pub const NET_CONNECT_TIMEOUT: Duration = Duration::from_millis(8 * 1000);
+pub const NET_CONNECT_TIMEOUT: Duration = Duration::from_millis(16 * 1000);
 
-pub const NET_IO_TIMEOUT: Duration = Duration::from_millis(8 * 1000);
+pub const NET_IO_TIMEOUT: Duration = Duration::from_millis(16 * 1000);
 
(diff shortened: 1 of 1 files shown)


jacksongoode:
Thanks for the PR! @Insprill do you think this would cause any side effects? I think it's within reason for long playlists.

psst-core/src/util.rs
```diff
@@ -5,9 +5,9 @@ use quick_protobuf::{BytesReader, MessageRead, MessageWrite, Writer};
 
 use crate::error::Error;
 
-pub const NET_CONNECT_TIMEOUT: Duration = Duration::from_millis(8 * 1000);
+pub const NET_CONNECT_TIMEOUT: Duration = Duration::from_millis(16 * 1000);
```
Insprill, line 8: Does the connection timeout need to be increased?

Insprill (changes requested):

arch-btw:
Hello, thank you both! 
Yes you're right, only NET_IO_TIMEOUT needs an increase for this to work, not the other one.

jacksongoode:
@Insprill I can remove the change from NET_CONNECT_TIMEOUT and we can merge?

Insprill:
Yeah that'd be good, I'd like to get this merged too.

jacksongoode:
See https://github.com/jpochyla/psst/pull/442

PR #448, opened 2024-01-03
Makes Sidebar/Main_view split resizable

Liberontissauri:
I use a tilling window manager and it throws me off how large the sidebar is because of the window occupying the whole screen. Makes the split between the sidebar and the main view draggable so it can be resized.

### DIFF
--- a/psst-gui/src/ui/mod.rs
+++ b/psst-gui/src/ui/mod.rs
@@ -158,7 +158,8 @@ fn root_widget() -> impl Widget<AppState> {
         .bar_size(1.0)
         .min_size(150.0, 300.0)
         .min_bar_area(1.0)
-        .solid_bar(true);
+        .solid_bar(true)
+        .draggable(true);
 
     ThemeScope::new(split)
         .controller(SessionController)


Liberontissauri:
Maybe adding a setting in the preferences to persist the ratio between app launches would be nice, although I'm not sure about how to implement that just yet

jacksongoode:
I question whether this would eventually lead to more layout bugs to resolve. If there is a way to set some min/max threshold of the side bar's resizing I think this could be acceptable. @Insprill any thoughts?

Liberontissauri:
Weirdly so, the sidebar already does have a minimum threshold for its size:

minimum_sidebar_size

I cant reduce it further. I'm not sure if this is part of druid itself, or if there's a way to change it. It does look like it exists to prevent overflow that makes content "invisible".

Liberontissauri:
Looks like it also applies to increasing the sidebar's size:
image

jacksongoode:
If you can set reasonable ratios that would keep the current min and make the max Not more than 1/3 of the width, I think that might resolve any issue with the sidebar being too wide as in your example.

jacksongoode:
@Liberontissauri Any followup with the suggested constraints?

PR #450, opened 2024-01-22
Adding support for Universal Binary for MacOS on M1

dscso:
I added support for universal binaries. The dmg contains an .app which runs native on aarch64 and x86 architecture. I tested it on my M1.
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,17 +55,17 @@ jobs:
         run: |
           rustup target add aarch64-apple-darwin
           cargo build --release --target aarch64-apple-darwin
-      
-      - name: Building universal Binay
-        if: ${{ runner.os == 'macOS' }}
-        run: lipo target/release/psst-gui target/aarch64-apple-darwin/release/psst-gui -create -output target/release/psst-gui
-    
+          
       - name: Bundle macOS Release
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -49,13 +49,23 @@ jobs:
...
+          cargo build --release --target aarch64-apple-darwin
+          
       - name: Bundle macOS Release
         if: ${{ runner.os == 'macOS' }}
         run: |
           cargo install cargo-bundle
           cargo bundle --release
         working-directory: psst-gui
+      
+      - name: replacing binary with universal binary
```
jacksongoode, line 66: If this could be "Create macOS universal binary" to fit the tense and use of lipo, I think it would be good to merge!

jacksongoode (changes requested):
👍

dscso:
If you want I can also change the name of the generated binary from `Psst-x64.dmg` -> `Psst.dmg`. I just was not sure if this would have side effects

Insprill:
That would be good. Check to see if it's linked in the readme, I think that would be the only thing that needs to be updated.

jacksongoode (approved):
I think we can merge, these linting issues are irrelevant to this PR.

PR #452, opened 2024-02-07
Resolve linting warnings

jacksongoode:
This resolves some of the linting errors that were causing it to break and replaces a .try_into() with just an .into().


jacksongoode:
@Insprill I think this is good to go, but just wanna make sure I'm not doing anything egregious.

Insprill (approved):
LGTM

PR #453, opened 2024-02-07
Adds ability to seek with arrow keys

jacksongoode:
This changes the arrow key behavior:
- Left/right - seek 10s forward/behind
- Shift+left/right - Previous/next track

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -412,11 +412,30 @@ where
                 ctx.set_handled();
             }
             Event::KeyDown(key) if key.code == Code::ArrowRight => {
-                self.next();
+                if key.mods.shift() {
+                    self.next();
+                } else {
+                    if let Some(now_playing) = &data.playback.now_playing {
+                        let current_position = now_playing.progress;
+                        let max_position = now_playing.item.duration();
+                        let seek_position =
+                            (current_position + Duration::from_secs(10)).min(max_position);
+                        self.seek(seek_position);
+                    }
+                }
                 ctx.set_handled();
             }
             Event::KeyDown(key) if key.code == Code::ArrowLeft => {
-                self.previous();
+                if key.mods.shift() {
+                    self.previous();
+                } else {
+                    if let Some(now_playing) = &data.playback.now_playing {
+                        let current_position = now_playing.progress;
+                        let seek_position =
+                            current_position.saturating_sub(Duration::from_secs(10));
+                        self.seek(seek_position);
+                    }
+                }
                 ctx.set_handled();
             }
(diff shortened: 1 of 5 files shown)


psst-gui/src/ui/playback.rs
```diff
@@ -90,29 +90,27 @@ fn playing_item_widget() -> impl Widget<NowPlaying> {
...
+                    .context_menu(|now_playing| match &now_playing.item {
+                        Playable::Track(track) => {
+                            track::track_menu(track, &now_playing.library, &now_playing.origin)
+                        }
+                        Playable::Episode(episode) => {
+                            episode::episode_menu(episode, &now_playing.library)
+                        }
+                    }),
+                1.0,
+            ),
```
jacksongoode, line 113: This was a linting format.

jacksongoode:
@Insprill If you have some time, let me know if this new functionality looks good to you.

psst-gui/src/controller/playback.rs
```diff
@@ -412,11 +412,25 @@ where
...
             }
             Event::KeyDown(key) if key.code == Code::ArrowRight => {
-                self.next();
+                if key.mods.shift() {
+                    self.next();
+                } else if let Some(now_playing) = &data.playback.now_playing {
+                    let current_position = now_playing.progress;
+                    let max_position = now_playing.item.duration();
+                    let seek_position =
+                        (current_position + Duration::from_secs(10)).min(max_position);
```
Insprill, line 421: Is there a particular reason this is implemented differently when going forward vs. backward?

Insprill, line 421: We should probably make the skip duration configurable - I can see many people wanting that.

Insprill (changes requested):

psst-gui/src/controller/playback.rs
jacksongoode: Good idea!

jacksongoode: Noted, recent commit standardizes this.

jacksongoode: Implemented in the recent commit.

psst-gui/src/ui/preferences.rs
```diff
@@ -212,6 +212,22 @@ fn general_tab_widget() -> impl Widget<AppState> {
 
     col = col.with_spacer(theme::grid(3.0));
 
+    col = col
+        .with_child(Label::new("Step Duration").with_font(theme::UI_FONT_MEDIUM))
```
Insprill, line 216: I think this should be called "Skip Duration" or "Seek Duration"

jacksongoode: Good call

Insprill (approved):
LGTM!

Issue #431, opened 2023-07-23
"Add to queue" in right-click menu


psharma04:
**Is your feature request related to a problem? Please describe.**
Currently, there is no equivalent of the "add to queue" button on the first party apps.

**Describe the solution you'd like**
When you right-click on a song, it should provide a "play next" or "add to queue" option.

**Describe alternatives you've considered**
N/A

**Additional context**
![in Psst](https://github.com/jpochyla/psst/assets/22587474/2d944576-9f3d-4233-b9dd-91ea7bdd55ed)
![in the MacOS Spotify App](https://github.com/jpochyla/psst/assets/22587474/c6be2974-8db4-4bfd-bdbe-9c0436ac2e2d)

SO9010:
@jacksongoode should we close this now?

jacksongoode:
Haha yeah, thanks for catching this!

PR #471, opened 2024-05-07
Add ability to add to queue

SO9010:
This adds the ability to right-click on a song and add it to play next in the queue. Please do leave me feedback and critiques as this is my first time programming in rust!

Solves #431 !

### DIFF
--- a/psst-core/src/player/mod.rs
+++ b/psst-core/src/player/mod.rs
@@ -119,6 +119,7 @@ impl Player {
             PlayerCommand::Seek { position } => self.seek(position),
             PlayerCommand::Configure { config } => self.configure(config),
             PlayerCommand::SetQueueBehavior { behavior } => self.queue.set_behaviour(behavior),
+            PlayerCommand::AddToQueue { item } => self.queue.add(item),
             PlayerCommand::SetVolume { volume } => self.set_volume(volume),
         }
     }
@@ -282,7 +283,7 @@ impl Player {
             loading_handle,
         };
     }
-
+    
     fn set_volume(&mut self, volume: f64) {
         self.audio_output_sink.set_volume(volume as f32);
     }
@@ -421,6 +422,9 @@ pub enum PlayerCommand {
     },
     SetQueueBehavior {
         behavior: QueueBehavior,
+    },
+    AddToQueue{
+        item: PlaybackItem,
     },
     /// Change playback volume to a value in 0.0..=1.0 range.
     SetVolume {
--- a/psst-core/src/player/mod.rs
+++ b/psst-core/src/player/mod.rs
@@ -283,7 +283,7 @@ impl Player {
             loading_handle,
         };
     }
-    
+
     fn set_volume(&mut self, volume: f64) {
         self.audio_output_sink.set_volume(volume as f32);
     }
@@ -423,7 +423,7 @@ pub enum PlayerCommand {
     SetQueueBehavior {
         behavior: QueueBehavior,
     },
-    AddToQueue{
+    AddToQueue {
         item: PlaybackItem,
     },
     /// Change playback volume to a value in 0.0..=1.0 range.
--- a/psst-core/src/player/queue.rs
+++ b/psst-core/src/player/queue.rs
@@ -54,10 +54,14 @@ impl Queue {
         self.user_added_items.push(item);
     }
 
-    fn handle_added_queue(&mut self){
+    fn handle_added_queue(&mut self) {
         if self.user_added_items.len() > self.user_added_items_position {
-            self.items.insert(self.positions.len(), self.user_added_items[self.user_added_items_position]);
-            self.positions.insert(self.position + 1, self.positions.len());
+            self.items.insert(
+                self.positions.len(),
+                self.user_added_items[self.user_added_items_position],
+            );
+            self.positions
+                .insert(self.position + 1, self.positions.len());
             self.user_added_items_position += 1;
         }
     }
--- a/psst-core/src/player/queue.rs
+++ b/psst-core/src/player/queue.rs
@@ -18,9 +18,9 @@ impl Default for QueueBehavior {
 
 pub struct Queue {
     items: Vec<PlaybackItem>,
-    added_items: Vec<PlaybackItem>,
+    user_added_items: Vec<PlaybackItem>,
     position: usize,
-    added_items_position: usize,
+    user_added_items_position: usize,
     positions: Vec<usize>,
     behavior: QueueBehavior,
 }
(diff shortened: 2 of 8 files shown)


jacksongoode:
This looks good but it sure would be nice to see the queue as a view. I'm wondering how difficult it would be to create a list page for a view of this queue? I could see us merging this feature before that, but it would feel incomplete... @Insprill what do you think?

SO9010:
I just realised that the queue isn't held when switching to another playlist or clicking on another song to play now so I will quickly fix that.

SO9010:
It will take me a decent amount of time to get it so it can keep the queued song so it can be played next in a playlist the song isn't in.

SO9010:
Okay, so I have added some updates to my code, the added queue is added to a new vector which means it can be held in place, additionally it means that I used the push_back method meaning that it doesn't make it play next, rather more like the official client it adds at the end of a user created queue... I hope that makes sense, but it is a lot better now.

jacksongoode:
Really excited to have a look at this today @SO9010 !

jacksongoode:
It seems like queuing, switching contexts (to a different playlist or album) and then adding again to the queue causes past queued items to disappear. Is this not the case in your behavior?

SO9010:
Hello, sorry this was an oversight! I assumed that the user would want to have the queue reset when they change the playlist. But that would be counter-intuitive. I have fixed this issue. 

I'm sorry if this is very time-consuming for you, I'm still pretty new to open-source contribution.

jacksongoode:
I just tested this out and it feels right to me! Looking over the code now!

psst-core/src/player/mod.rs
```diff
@@ -282,7 +283,7 @@ impl Player {
             loading_handle,
         };
     }
-
+
```
jacksongoode, line 286: Make sure to lint :)

psst-core/src/player/queue.rs
```diff
@@ -18,7 +18,9 @@ impl Default for QueueBehavior {
 
 pub struct Queue {
     items: Vec<PlaybackItem>,
+    added_items: Vec<PlaybackItem>,
```
jacksongoode, line 21: Maybe there is a better name for this var? I'm not sure what it would be but `added_items` vs `items` are not clearly distinct in their usage from the names alone.

psst-core/src/player/queue.rs
```diff
@@ -46,6 +50,23 @@ impl Queue {
...
+            self.items.insert(self.positions.len(), self.added_items[self.added_items_position]);
+            self.positions.insert(self.position + 1, self.positions.len());
+            self.added_items_position += 1;
+            
+            if self.added_items_position > 2 {
+                self.added_items.remove(0);
+                self.added_items_position -= 1;
+        }
+        }
+    }
```
jacksongoode, line 68: This logic is a little bit messy, could you describe what needs to happen here? Maybe there is a cleaner way of implementing this?

psst-gui/src/controller/playback.rs
```diff
@@ -342,7 +348,8 @@ where
                     data.start_playback(queued.item, queued.origin, progress.to_owned());
                     self.update_media_control_playback(&data.playback);
                     self.update_media_control_metadata(&data.playback);
-                } else {
+                }
```
jacksongoode, line 351: Linting fix here too

psst-gui/src/data/mod.rs
```diff
@@ -180,11 +182,19 @@ impl AppState {
...
-            .find(|queued| queued.item.id() == item_id)
-            .cloned()
+        if let Some(queued) = self.playback.queue.iter().find(|queued| queued.item.id() == item_id).cloned() {
+            return Some(queued);
+        } else if let Some(queued) = self.added_queue.iter().find(|queued| queued.item.id() == item_id).cloned() {
+            return Some(queued);
+        } else {
+            None
+        }
+    }
```
jacksongoode, line 192: Hmm, can you justify why we keep a separate queue from the contextual playback queue? I just maybe want to think about the justification.

jacksongoode (changes requested):
Looks good for the most part! Just some linting cleanup that needs to happen (use the standard linter) and just some clarification on the arch decisions.

psst-core/src/player/queue.rs
SO9010: Do you think that user_items would be clearer?

psst-core/src/player/queue.rs
SO9010: My explanation for my code: The initial if statement checks if the position of the user-added queue has caught up with its length. Once that happens, the item is added to the main queue, and its position is updated in the main queue's positions.

Then, the position of the added item is incremented.

The next part ensures the vector doesn't get too long, but this isn't typically handled within the program, so I could probably remove that.

psst-gui/src/data/mod.rs
SO9010: I believe that having a separate queue allows the program to avoid removing the user queued songs. When the user clicks on a different playlist, a new queue is created entirely meaning the main song queue gets rewritten, in my opinion showing the need for a separate queue.

This approach also ensures that the song contexts are passed correctly. Otherwise, an error occurs where only the audio loads, but none of the data is found in the main fetched queue.

From a UI perspective, having a separate user-added queue allows for a more minimal display of just the user-added songs. Alternatively, it provides the option to have two different scrollable contexts where the user can view both queues separately.

Please let me know if this made sense and thank you so much for your feedback :)

psst-core/src/player/queue.rs
jacksongoode: Can we assume that adding to the queue will always increase its length? Is that first check needed?

psst-gui/src/data/mod.rs
jacksongoode: Thanks for the explanation, those were my thoughts as well just wanted some clarity on the arch decision for this PR.

psst-core/src/player/queue.rs
SO9010: The first check is necessary to verify whether the user has gone through their added songs. The variable added_items_position tracks which song needs to be added next. Without this check, it would continuously attempt to add a song to the main queue, even if there isn't one available.

Sorry if the first explanation didn't make much sense, I hope this helps.

psst-core/src/player/queue.rs
SO9010: I will change it to user_added_items

jacksongoode: User items works as well

jacksongoode (approved):
Alright, looks good!

SO9010:
Thank you so much!!!!

Issue #473, opened 2024-05-18
[Linux] debian/ubuntu package naming conflict


ChrisLauinger77:
**Describe the bug**
debian package naming conflict
https://packages.debian.org/trixie/psst

I am not an ubuntu user but they might have the same problem
https://launchpad.net/ubuntu/+source/psst

**To Reproduce**
install the deb from here and then apt update && apt upgrade

**Expected behavior**
debian should not replace the package with the one from the repo

**Environment**
 - OS: debian
 - Version: any

**Additional context**
maybe rename the package to psst-gui or something ?

Insprill:
Duplicate of #391

Issue #391, opened 2023-04-21
Debian/Ubuntu/Mint: conflict with psst package


tobiasrenkin:
**Describe the bug**
There is an unrelated Debian package called [psst](https://packages.debian.org/sid/devel/psst). If I install the psst spotify client using the .deb package provided [here](https://nightly.link/jpochyla/psst/workflows/build/master) and run apt-get upgrade, the psst spotify client package is "upgraded" to the completely unrelated debian "psst" package, i.e. effectively uninstalled and replaced.

I don't need the other package and I can mark psst the spotify client as hold to solve this. But other people might want to have both of them installed, so I would suggest to rename the debian package to psst-spotify or something similar.

**To Reproduce**
See above.

**Expected behavior**
The two packages co-exist.

**Environment**
 - OS: Linux Mint
 - Version: 21.1

**Additional context**
Should affect all debian-based distros.

Insprill:
I'm not sure what can be done about this, if anyone knows more or has any ideas please share them!

cloudybyte:
I believe the only real solution is to rename the package to either `psst-spotify`, `psst-player` or sth.

brianclinkenbeard:
As a temporary solution, you can use `sudo apt-mark hold psst`. Or, if you don't want to be notified that psst is being held back, you can pin the package by writing the following to `/etc/apt/preferences.d/psst-pin`:
```
Package: psst
Pin: version 1
Pin-Priority: 1001
```

ChrisLauinger77:
I just would name the package:
psst-gui

PR #475, opened 2024-05-18
Linux desktop file and debian package name

ChrisLauinger77:
I also added the svg icon
which could be deployed to /usr/share/icons
but I don't know what to edit to achieve this :(
This PR addresses #391 & #473


ChrisLauinger77:
Found the place in build.yml meanwhile - svg will be copied there

.pkg/DEBIAN/control
```diff
@@ -1,4 +1,4 @@
-Package: psst
+Package: psst-gui
```
jacksongoode, line 1: Is there a reason to rename this package?

.github/workflows/build.yml
```diff
@@ -126,6 +126,8 @@ jobs:
             mkdir -p "${{runner.workspace}}/pkg/usr/share/icons/hicolor/${LOGO_SIZE}x${LOGO_SIZE}/"
             cp "./psst-gui/assets/${LOGO}" "$_/psst.png"
           done
+          mkdir -p "${{runner.workspace}}/pkg/usr/share/icons/hicolor/scalable/apps/"
+          cp "./psst-gui/assets/psst.svg" $_
```
jacksongoode, line 130: Could you call the SVG `logo.svg` to match the other assets included?

jacksongoode (changes requested):
Looks good for the most part

.pkg/DEBIAN/control
ChrisLauinger77: Yea #391

Insprill: https://github.com/jpochyla/psst/issues/391

.github/workflows/build.yml
ChrisLauinger77: Done

Issue #255, opened 2022-01-11
Notifications Won't Disappear


ghost:
**Describe the bug**
These Notifications won't disappear sometimes.

**To Reproduce**
Just add a bunch of songs to a playlist. Some notifications will disappear, others not.

**Expected behavior**
They should all disappear.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/12158101/148930726-a8334e99-5aaa-42de-9d7a-2b3ef697681c.png)

**Environment**
 - OS: Windows 10

jacksongoode:
Fixed with #503

PR #478, opened 2024-06-12
Add Close Button to Notifications and Improve Cosmetic Appearance

SO9010:
This merge request partially addresses issue #255, which I believe stems from an upstream issue. After exploring several solutions, including making the notification close when the user hovers over it, the problem persisted. I thoroughly reviewed all the code related to time handling to ensure that wasn't the problem.
Changes Made:

Added Close Button: Introduced a close button to the notification for more explicit user control.

Cosmetic Improvement: Made the notification slightly transparent for a better visual appeal.

These changes should enhance user experience by providing a clear way to dismiss notifications and a more polished look.

Here is a screenshot of it:

image

Please review and merge if everything looks good. Thank you!


jacksongoode:
This is nice, thank you!

However, if this is to address a bug where notifications are not being cleared, I think it might be better to trigger some sort of notification clear method which gets rid of all the notifications. This could be as simple as pressing escape and triggering all the notifications to clear. I think it might be a little bit redundant to add the option to clear each individual notification given that they are intended to be ephemeral, temporary status messages. So I would advocate for a global clear rather than this individual UI change.

But even better would be to see why the the notifications are persisting in the first place.

jacksongoode:
We could keep this or close it now that https://github.com/jpochyla/psst/pull/503 is ready.

SO9010:
I'll close it as it was only a temporary solution, as I thought it was an upstream issue.

PR #481, opened 2024-06-17
Fix macOS preferences window contols and icon from side-panel

jacksongoode:
This removes the old icon from the side panel as it was taking up a lot of space. It also fixes the window controls that were being overlapped on macOS.

Before:
image

After:
image

image


Insprill (approved):
LGTM. We should fix that failing clippy lint so CI stops failing (can be done in a separate PR).

PR #482, opened 2024-06-18
Fix linting & code style

jacksongoode:
This also removes the odd left padding with the max loaded tracks and seek duration, which is inconsistent with the rest of the UI.


psst-gui/src/ui/preferences.rs
```diff
@@ -228,7 +228,6 @@ fn general_tab_widget() -> impl Widget<AppState> {
                         |usize: &usize| usize.to_string(),
                     )),
                 )
-                .padding((theme::grid(1.5), 0.0))
```
Insprill, line 231: Were these lints or are they fixing a UI issue?

jacksongoode: Just fixing a UI issue, they really shouldn't have been there given that we don't left pad anything else.

PR #484, opened 2024-06-26
Fix macOS universal build & support macOS >= 11

jacksongoode:
This fixes an issue with our universal builds, it also:
- adds the `MACOSX_DEPLOYMENT_TARGET` so we can support older versions than what the macOS runner is set to
- Lints a few depreciated `usize` imports
- Bumps the version of the actions to use Node 20 so we don't get depreciation warnings

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,23 +51,15 @@ jobs:
         if: runner.os != 'macOS'
         run: cargo build --release
 
-      - name: Build for macOS (x86_64 and aarch64)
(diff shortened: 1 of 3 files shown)


jacksongoode:
@alejandrot96 I think this is good to go, but I will need someone to test and confirm this.

jacksongoode:
I had someone on Intel running 11.6 confirm this works on their machine.

alejandrot96:
last one is working on ventura 👌🏽

Issue #476, opened 2024-05-21
New layout proposition.


SO9010:
**Is your feature request related to a problem? Please describe.**
The layout isn't great and it could be improved.

**Describe the solution you'd like**
I think that a new side bar could be added which contains the playback queue. There would be three sections here:

One main section with the curently playing song with the album cover and the information with it, this could partially replace the currently playing section in the currernt design. 

One section for the user added queue.

One section for the queue that will be played from the playlist.

I think that the proportions could be inproved. the play back bar isn't in-line with the bar on the right. Same with the top. 

I think that the volume bar can be moved to the right of the playback options. To replace where it originally was we can put the search bar. 

Then I think that there coud be a playlist pinning option where the search bar used to be.

Additionally, optionally I think that there could be a search icon at the top right of the playlist to search the playlist.

I think the muted mic icon is a bit pointless and should be replaced by the Psst icon. 

**Please refer to my mock-up for an example**

**Describe alternatives you've considered**
Yes, the search icon in the top right is not nessisary.
I am happy to work on this in my free time, I just want to make sure that my idea alligns in a similar way to all of yours!

**Additional context**
![psst](https://github.com/jpochyla/psst/assets/77629938/a8f9a769-31c0-4359-90e3-317bd760f569)

jacksongoode:
I definitely have been thinking about how we ought to display the queue now that functionality is provided.

- I think the three panel approach that Spotify adopted could work. I worry a bit about including the album art there given this is intended to be a minimal player. If we want to take this approach it should definitely be collapsible.
- Agree that the muted mic icon is dated and takes up valuable space. I honestly might advocate to remove it entirely and use the space for playlists/right sidebar.
- Disagree on putting any album artwork in the left side panel, I think that space is too valuable.
- I think the search bar should still be kept at the top of the sidebar given its priority.

SO9010:
Perhaps just a collapsible scroll view on the right (much like shown) could be used where it is automatically shown when the window is over a certain size, and if not it can be temporarily shown using a button?

jacksongoode:
That seems good, I would also think that we should lock the width of the left and right sidebars. Currently, Spotify's desktop is a mess with how it has to handle resizing. It seems like a headache to try to handle all the cases. We can just deal with the reflow of the center and disable the right sidebar if its too small of a window size maybe?

SO9010:
[...]

Issue #477, opened 2024-06-10
Removes multiple songs if there are duplicates


SO9010:
**Describe the bug**
When having a song added more than once in a playlist, then you remove one, it removed every single song which is the same. This makes it annoying to then re-add the song, especially if you forget which song this is.

**To Reproduce**
Add a song multiple times to a playlist, then right click onto the song, and then observe all of the songs that are the same disappearing.

**Expected behavior**
Remove just the one song.

**Screenshots**

**Environment**
 - OS: Arch linux

jacksongoode:
This is definitely a bug. We should identify the song key not just by the song itself but also perhaps its position in the playlist.

jacksongoode:
Figured it out. It's actually undocumented but you can remove tracks by index.

https://github.com/JMPerez/spotify-dedup/commit/d0493aa0b760a3e4dd5d747b7cae605fbec60c9c

PR #485, opened 2024-06-30
Layout Change: Queue View Widget

SO9010:
This is addressing #476 on creating a new layout and this is a draft PR so that  @jacksongoode is able to collaborate with me.

### Features: 
- [x] Fix sizing on the left and right panel 
- [x] A sidebar on the right which hides with depending on the window size
- [x] A new divider for the left and right sides as that was brought in by the split view which I changed to be a flex row so then the left and right bars don't take up too much space
- [x] Queue handling functions
       - Clear queue
       - Remove from queue

### Questions:
- Is it worth aligning some of the dividers for aesthetic purposes?
- Should we have a function which skips to the place in the user queue where the user clicks, I would advocate for this?
- What should we do after the song has been played from the queue, should it be removed from the view, or should we implement some sort of auto scroll?
- I think we should have an option to disable this queue somewhere, maybe in settings or maybe in the top right of the title bar like here: 
image

---
### Current look of this PR: 

https://github.com/jpochyla/psst/assets/77629938/e1d097dc-89ab-4cdb-8585-583db7b69625


jacksongoode:
I'll have a look at this tonight! Very excited!

jacksongoode:
image

Just made a few cleanups just to get consistency, however, it might be better to use reuse objects rather than recreate them.

jacksongoode:
@SO9010 Hey Sam, just curious if you've had the time to look over the visual changes so far?

SO9010:
Hey @jacksongoode, sadly I don't have access to my computer! Your changes look great to me. That bug where the separator is off by 1px is very annoying! I will take a look when can and try to maybe hack a fix for it. But it still looks a ton better than with just a small separator bar.

Also, I'm thinking of implementing some sort of hover function to the close button to make it dimmer when not needed as it seems a bit intrusive and bulky right now. 

I think the Ul is looking really good! I'm happy to take any suggestions. :)

SO9010:
Okay so an update. I have been working on this today and have come up with a solution for the off by one pixel issue and that was to fix the height. Additionally I added the ability to add dividers between the left and right sides! I also aligned the bottom player with the left controller section!


image


However, I am a bit stuck on how I get the position of the song in the queue to get removed after the button has been hit! Any suggestions @jacksongoode ??

psst-gui/src/ui/mod.rs
```diff
@@ -129,13 +128,14 @@ fn root_widget() -> impl Widget<AppState> {
...
-        .fix_height(100.0)
+        .fix_height(88.0)
         .background(Border::Top.with_color(theme::GREY_500));
 
     let sidebar = Flex::column()
         .with_flex_child(playlists, 1.0)
         .with_child(controls)
-        .background(theme::BACKGROUND_DARK);
+        .background(Border::Right.with_color(theme::GREY_500))
+        .fix_width(185.0);
```
jacksongoode, line 138: I think it would still be nice to have this dynamically resize. Is there something preventing this?

psst-gui/src/widget/utils.rs
```diff
@@ -181,11 +185,31 @@ impl Border {
...
+            };
+            let y2 = match self {
                 Self::Top => h / 2.0,
                 Self::Bottom => ctx.size().height - h / 2.0,
+                Self::Left => h / 2.0,
+                Self::Right => h / 2.0,
             };
-            let line = Line::new((0.0, y), (ctx.size().width, y));
+            let line = Line::new((w1, y1), (w2, y2));
             ctx.stroke(line, &color.resolve(env), h);
```
jacksongoode, line 213: Thanks for adding this!

psst-gui/src/ui/mod.rs
```diff
@@ -150,18 +150,18 @@ fn root_widget() -> impl Widget<AppState> {
...
-        .bar_size(1.0)
-        .min_size(150.0, 300.0)
-        .min_bar_area(1.0)
-        .solid_bar(true);
+        
+    let split = Flex::row()
+        .with_child(sidebar)
+        .with_flex_child(main, 1.0)
+        .with_child(queued::queue_widget())
+        .background(theme::BACKGROUND_DARK);
```
jacksongoode, line 158: Ah, so the side bar (with playlists) now doesn't follow a split ratio. Was there a reason that this logic was implemented as a fix or more so a preference?

SO9010:
I thought it helped with consistency and ensuring that it doesn't become unreasonably large. Feel free to change it!On 14 Jul 2024 13:18, Jackson ***@***.***> wrote:
@jacksongoode commented on this pull request.



In psst-gui/src/ui/mod.rs:
> +    let split = Flex::row()
+        .with_child(sidebar)
+        .with_flex_child(main, 1.0)
+        .with_child(queued::queue_widget())
+        .background(theme::BACKGROUND_DARK);

Ah, so the side bar (with playlists) now doesn't follow a split ratio. Was there a reason that this logic was implemented as a fix or more so a preference?

psst-gui/src/ui/mod.rs
SO9010: I ran into difficulties with having it dynamically resize due to the split function still giving space to have a widget on the right even when the queue widget has disappeared.

psst-gui/src/ui/mod.rs
SO9010: It was to follow consistency with the right-hand side as that currently doesn't move, and additionally to make it so the sidebar doesn't become unreasonably wide.

psst-gui/src/ui/mod.rs
SO9010: I just realised that I just need to not use the split widget and I can just use the flex child! I will change now!

psst-gui/src/ui/mod.rs
SO9010: I have now changed it to follow the split ratio!

psst-gui/src/ui/mod.rs
jacksongoode: Right! Good sleuthing!

jacksongoode:
image

@SO9010 Just noticed this little overlap with the widgets of the section selectors overlaying the border.

psst-gui/src/ui/mod.rs
```diff
@@ -150,18 +149,18 @@ fn root_widget() -> impl Widget<AppState> {
         .with_flex_child(Overlay::bottom(route_widget(), alert_widget()), 1.0)
         .with_child(playback::panel_widget())
         .background(theme::BACKGROUND_LIGHT);
-
-    let split = Split::columns(sidebar, main)
-        .split_point(0.2)
-        .bar_size(1.0)
-        .min_size(150.0, 300.0)
-        .min_bar_area(1.0)
-        .solid_bar(true);
```
jacksongoode, line 159: I'm wondering if maybe if it wouldn't be worthwhile trying to use two Split:columns() one for `let split = Split::columns(sidebar, main)` and then one for the `Split::columns(split, queue_widget)`? Maybe this would allow the split borders to exist nicely? Just a thought

SO9010: This would only work for the left and not the queue widget. This is because it will still leave remaining space for the widget even if it is empty.

But it would work to stop the overlapping on the left which isn't an issue in the right so this would work if we only use the split widget for the left sidebar. 

Does that make sense?

SO9010:
I have fixed the overlapping issue by reintroducing the split widget between the left sidebar and the main section. Additionally, I have created a new section called the right sidebar. Now, we have both a left sidebar and a right sidebar. This change ensures that the queue widget no longer has problems with disappearing, which occurred when it was dynamically resizeable.

I have a question now, how should we deal with it when a song has been played from the queue? Should we remove it from the widget, or should we have it, so the user is able to essentially skip backwards?

SO9010:
I have made it so the user can essentially skip backwards in the queue, as that feels quite fluid.

SO9010:
Hi @jacksongoode,

Could you please take a look at the program? I'd like to outline a few things:

The reason for not reusing the widget for the tracks is that we need to have an iterator method. I found this a bit complex to do, so I think it's best to just leave it as a new widget, you are welcome to change this of course.

There is currently an issue that is very similar to issue #477. I need some help with this because I can't figure out how to single out the position of a track in the list view, as right now it just removes the first song with the same ID, but we could remove it using the index.

I'm thinking that we need some sort of indication for which song from the queue is currently playing. What do you think?

I have made it so that the queue hides if there are no items added by the user, which helps maintain the aesthetics of the program nicely.

There is also an issue of what happens when a user clicks on the queue widget before playing anything from a playlist, which I still need to solve!

Here is a video of how it currently looks: 

https://github.com/user-attachments/assets/e938f076-a278-4587-bb55-4165090e2098

SO9010:
Yes makes sense!

jacksongoode:
> There is currently an issue that is very similar to issue https://github.com/jpochyla/psst/issues/477. I need some help with this because I can't figure out how to single out the position of a track in the list view, as right now it just removes the first song with the same ID, but we could remove it using the index.

I do think we should probably fix this as it's own issue. The way's ID's should exist should contain their index in the list.

SO9010:
Ok makes sense, I think I'm starting to understand the issue more so will be able to work on it after this has merged!

jacksongoode:
https://github.com/jpochyla/psst/pull/499 This should fix the duplicate ID track issue.

jacksongoode:
Now that https://github.com/jpochyla/psst/pull/499 is merged, we can rebase and the fix should be here.

SO9010:
I had to implement it slightly differently because I was unable to reuse the normal view widget, which has to take an iterable track. The queue works through a vector of QueueEntry's. However, I have started implementing a similar way of collecting the song index, which is working very well :)

SO9010:
Okay, here's a pleasing update: the end is near! I have implemented the remove by index, which has worked very well and removed the big issue, and now I have implemented the method which skips to a certain part of the 

There are very few things I have to do: remove the first song after it has been played, trigger audio playback when a song is clicked from the queue, and add a little section in the settings to keep the queue view hidden.

SO9010:
@jacksongoode, how should we do it with clearing the queue? I have a button to the right of the title, but it looks wrong. What do you suggest?

SO9010:
Yes, but I have quickly added a few bits to the context menu!

SO9010:
There is, in fact, another issue!! When the user clicks on the first item in the queue, it doesn't do anything!

SO9010:
Alright, @jacksongoode, I have a question! Do you think we should do something slightly different to how it currently functions and follow Spotify more? Currently we have two places where it shows the currently playing song, at the top of the queue, and in the playcontrol section, but this doesn't need to be. We could remove it as it gets played, as the user will see it going to the now-playing section at the bottom with all the controls.

jacksongoode:
Right, I noticed that! Yeah I do think we should pop the queued item the moment it gets played - that's how Spotify does it. So when a new queued item gets promoted to play, we immediately remove it from the queue list.

SO9010:
Ok, so progress again today; I have managed to make it so the first song gets popped after it played. Now, all that is needed is to initiate playback from the queue!

jacksongoode:
With these recent changes, it seems like the cue that gets added to the sidebar, even after being exhausted, does not reflect the internal cue. It doesn't seem like the internal cue gets removed in the same way.

SO9010:
Okay so could you possibly check this out for me :P thank you

jacksongoode:
It seems like there might be a bug around how the items in a queue get triggered when they change from a selected item into the queue. I get a really long spinning wheel and the track plays, but I think that indicates the method that triggered the playback might be wrong.

image

To reproduce:
- Open
- Queue a song from context #1
- Play a random song in a different context #2
- Click next or finish song, queued item will have spinning wheel

SO9010:
@jacksongoode I think I have fixed this; I had to introduce a new vector, the display vector, which shows only the queued items left to play and another one, which is almost like a catalogue for it to look through.

The initial issue was that I was removing the song while the program still needed it.

SO9010:
@jacksongoode This should be good now. Could you please check it for me?

I found the issue. It came from the fact I was deleting the first song from the queue when it was being played, but it needs this to know where the playback origin is, so I added a display vector for this.

jacksongoode:
Checking it out now! Thanks!

jacksongoode:
Ugh unfortunately I cant login since my config was delete and Spotify has revoked the user/pass method. So I'll need to make a PR to address that.

jacksongoode:
Ok I'm halfway done with that auth PR, just enough to get a good token.

So I see two issues here:

1. Queuing a song and then click on that item at the top of the queue should just pop the song and play it. Currently only clicking on items after the first will trigger them to play.

2. Clicking pops wrong song:
- Queue 3 songs
- Play a random song to start playback
- Skip 3 songs
- Queue 3 more songs
- Click 2nd queued item from top
- See that song played is not the clicked track

SO9010:
Hello, thank you so much for checking it out! These issues should be relatively easy to sort out! :) I'll let you know when I've sorted that out. Sometimes these bugs slip as I've been looking at the same thing for a while :)

jacksongoode:
I think this PR may need to be rebased but I could be wrong - I recently merged a fix to the builds and the oauth implementation.

SO9010:
Ok so Ive had an other look at this, in my opinion it is almost done, just that there is an issue where when removing the song that is currently playing, it removes it if someone clicks on the song that is about to play then it registers as the song that was queued.:

jacksongoode:
> it removes it if someone clicks on the song that is about to play then it registers as the song that was queued:

Oh I think the solution here is the context. Queued items should have some position or label that makes it clear that on clicking on items that are present in the queue should remove those queued items? It seems like a design bug if clicking anywhere else has any interaction with the queue's poping off the stack.

SO9010:
Heya, thanks for this; I don't think I get the bug where clicking anywhere else pops the top of the stack. Could you please send a video or something of this happening?

SO9010:
Yes I agree but annoyingly there will probably need to be some big refractoring needed as currently the way that the origin of a currently playing song is done through this function:
```
pub fn queued_entry(&self, item_id: ItemId) -> Option<QueueEntry> {
        
        if let Some(queued) = self
            .added_queue
            .iter()
            .find(|queued| queued.item.id() == item_id)
            .cloned()
        {
            Some(queued)
        } else if let Some(queued) = self
            .playback
            .queue
            .iter()
            .find(|queued| queued.item.id() == item_id)
            .cloned()
        {
            return Some(queued);
        } else {
            None
        }
    }
```

I added the top bit, which forces it to assume that if the items are in the added queue when that song ID is playing, it must be from the queue. However, this is where the issue comes in. I think I could possibly change this but this is quite core logic so id like to avoid completely changing this function. 

I've been looking at the idea of having a variable which says whether it is playing from the queue, much like I have done with the actual queue code. But I'm struggling with the implementation. 

Is there maybe a way to get a function to parse through a value in things like this section:
```
fn handle_command(&mut self, cmd: PlayerCommand) {
        match cmd {
            PlayerCommand::LoadQueue { items, position } => self.load_queue(items, position),
            PlayerCommand::LoadAndPlay { item } => self.load_and_play(item),
            PlayerCommand::Preload { item } => self.preload(item),
            PlayerCommand::Pause => self.pause(),
            PlayerCommand::Resume => self.resume(),
            PlayerCommand::PauseOrResume => self.pause_or_resume(),
            PlayerCommand::Previous => self.previous(),
            PlayerCommand::Next => self.next(),
            PlayerCommand::Stop => self.stop(),
            PlayerCommand::Seek { position } => self.seek(position),
            PlayerCommand::Configure { config } => self.configure(config),
            PlayerCommand::SetQueueBehavior { behavior } => self.queue.set_behaviour(behavior),
            PlayerCommand::SkipToPlaceInQueue { item } => self.queue.skip_to_place_in_queue(item),
            PlayerCommand::ClearQueue => self.queue.clear_user_items(),
            PlayerCommand::AddToQueue { item } => self.queue.add(item),
            PlayerCommand::RemoveFromQueue { item } => self.queue.remove(item),
            PlayerCommand::SetVolume { volume } => self.set_volume(volume),
        }
    }
```

As then for example we could have a player command GetIfPlayingUserItem, then from there, this would all be streamlined and could even be placed in the Appdata section to be more easily accessible.

jacksongoode:
I think this needs to be on a branch? It seem like there's some weird conflict I'm getting by this being on your master branch? Generally every PR should be on a branch with its name.

SO9010:
So, before I think this is possible, I think we need to find a way to give the playback origin without just having it detected after it's been played.

jacksongoode:
@SO9010 I would love to merge this, but I did test this a bit last night. I think there is still some issues around adding to queue and switching context, adding more items, and then clicking on an item in the queue and getting the wrong track playing. Could you see if that's reproducible on your end?

SO9010:
I can't lie I started to give up on it, but I will have a look soon if I can't get it in a better working order :) I would also love it to be part of the program.

Issue #443, opened 2023-10-10
feature: close preference window with Esc


WhyNotHugo:
**Is your feature request related to a problem? Please describe.**

When the preferences window is open, it can only be closed via the desktop-specific keyboard shortcut to close a window.

**Describe the solution you'd like**

Being able to quickly dismiss it with <kbd>Esc</kbd> would be quick and handy. It's a shortcut that I intuitively expected to work.

**Describe alternatives you've considered**

I guess an "okay" button could be used? It might be ambiguous given that preferences seem to be saved automatically and don't need a separate action to confirm.

**Additional context**

n/a

PR #486, opened 2024-07-02
Close preferences with Esc

jacksongoode:
Very simple, close preferences with the escape key. This resolves https://github.com/jpochyla/psst/issues/443 but it would really be great to finish this other issue as well https://github.com/jpochyla/psst/issues/103 but I think it requires thinking about how background play might work.


jacksongoode:
I'm just going to merge this given that there's very little I can imagine that would break with this.

PR #487, opened 2024-07-02
Bump all packages to latest versions

jacksongoode:
So far in my testing, things seem ok! I moved to use the zbus backend for souvlaki which should be a more minimal and correct implementation on Linux.


jacksongoode:
@SO9010 Would you be able to check this build on Linux and see if there are any obvious behavioral, CPU, memory or etc differences between main?

arch-btw:
I did this manually ever since last week and it works! No issues, it even seems like it's more responsive.

jacksongoode:
@arch-btw What platform are you on just out of curiosity?

arch-btw:
@jacksongoode I use Arch Linux 👍

PR #489, opened 2024-07-03
Arm build for Linux

jacksongoode:


Insprill:
I'd suggest we use [cross](https://github.com/cross-rs/cross) to build for multiple architectures.

jacksongoode:
> I'd suggest we use [cross](https://github.com/cross-rs/cross) to build for multiple architectures.

Thanks, I'll give it a go, this initial PR was just a stab in the dark.

jacksongoode:
@Insprill If you have any experience with `cross`, I'd be happy for your help. It seems there are some issues with packages needed to be installed for each architecture.

.github/workflows/build.yml
```diff
@@ -41,24 +48,28 @@ jobs:
       - name: Setup Cache
         uses: Swatinem/rust-cache@v2
 
-      - name: Install Linux Dependencies
-        if: runner.os == 'Linux'
-        run: sudo apt-get update && sudo apt-get install -y libgtk-3-dev libssl-dev libasound2-dev
-
```
Insprill, line 47: This is still required

jacksongoode: I suppose this is because `cross` uses the native action runner to do the amd compilation but not arm? I thought I could install these in the root `Cargo.toml` with:

```
[workspace.metadata.cross.target.aarch64-unknown-linux-gnu]
pre-build = [
    "dpkg --add-architecture $CROSS_DEB_ARCH",
    "apt-get update && apt-get --assume-yes install libgtk-3-dev:$CROSS_DEB_ARCH libssl-dev..."
]
```

But I get this error on the arm Ubuntu indicating they might not be installed...

```
cflags" "glib-2.0" "glib-2.0 >= 2.56"` did not exit successfully: exit status: 1
  error: could not find system library 'glib-2.0' required by the 'glib-sys' crate

  --- stderr
  Package glib-2.0 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `glib-2.0.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'glib-2.0' found
  Package glib-2.0 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `glib-2.0.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'glib-2.0' found
```

jacksongoode:
@Insprill Took long enough... anyways let me know if you have any suggestions with the current implementation.

jacksongoode:
@Insprill I'm going to go ahead since I can't see any major consequence. We can add back the test if we'd like in a later PR.

PR #491, opened 2024-07-04
Add alert message for when tracks are added to the queue

jholdnak:
This adds the ability to tell the user that the song of their choice was added to the queue. Feel free to leave me some feedback as this is my first time programming in Rust, and making an open-source contribution.

### DIFF
--- a/psst-gui/src/ui/playback.rs
+++ b/psst-gui/src/ui/playback.rs
@@ -10,13 +10,13 @@ use druid::{
 use itertools::Itertools;
 
 use crate::{
-    cmd,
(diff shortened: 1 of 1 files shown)


jacksongoode (approved):
This looks great! Thank you for contributing - and especially as it's your first time contributing to an OSS project!

jacksongoode:
@jholdnak If you want an even trickier challenge, you'll find that some of the alert messages don't disappear when added in succession. I think there's some issue with the debounce time. Each notification needs a timer that will cause it to vanish after it appears.

PR #495, opened 2024-07-09
Cache album art

jacksongoode:
Currently, this stores the album art work by hashed URI.

Ideally, we would be storing and retrieving the album art by the album ID. However, that would mean modifying the function in a way that I think would warrant its own PR. With that extension, we would only need to download album artwork for each album rather than album art for each track, which is happening now.

I think the increase in cache size is negligible here, given that each album cover as a small thumbnail is about 2-6kb and has a large thumbnail about ~40kb.

This will reduce the number of requests,  improve responsiveness and CPU likely.

### DIFF
--- a/psst-gui/src/webapi/cache.rs
+++ b/psst-gui/src/webapi/cache.rs
@@ -1,9 +1,12 @@
 use std::{
+    collections::hash_map::DefaultHasher,
     fs::{self, File},
+    hash::{Hash, Hasher},
     path::PathBuf,
     sync::Arc,
 };
 
+use druid::image;
 use druid::ImageBuf;
 use lru_cache::LruCache;
 use parking_lot::Mutex;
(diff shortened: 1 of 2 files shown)


Insprill:
I think this cache could add up quickly for people with large libraries. It might be worth bringing in the image crate to store the images in a more efficient format. It might be a little complex, but having a limit on size would probably be good too.

jacksongoode:
@Insprill Hmm, right now having loaded 2,200 small thumbnails, it only totals around 10mb but I would definitely be curious about how we could compress them further. Do you have any ideas? I suppose we could group them into blocks and then compress the blocks and save the index? There's not much compression gain on such small images individually.

jacksongoode:
I think additional improvements can be made in a subsequent PR? What do you think @Insprill? I'd rather merge this, see if it actually does pile up in some noticeable way and then make changes after.

Insprill (approved):
LGTM. 10mb isn't bad, I agree we can tweak later.

Issue #457, opened 2024-02-18
Clicking on save icon shouldn't trigger play on track


jacksongoode:
Currently there's no way to suppress the play back on click when the save unsaved action takes place.

PR #498, opened 2024-07-11
UI cleanup for save track icon

jacksongoode:
This PR:
- Fix some minor string capitalization consistency issues
- Switches the heart to an add icon (consistent with Spotify)
- Adds the icon to the row below the duration (to save save horizontal space and prevent reflow when tracks load)

Here's what it would look like:

image

image

However, this introduces an issue where clicking on the save icon actually triggers the parents play action - so that would need to be fixed.


SO9010:
> However, this introduces an issue where clicking on the save icon actually triggers the parents play action - so that would need to be fixed.

Isn't this the same issue as #457 so not newly introduced?

jacksongoode:
Yeah actually I made that issue thinking it was the same behavior with the standard UI but the regular UI doesn't put the saved icon on top of the clickable track widget - it adds it to the side.

Currently there are two issues with this PR:
- Preventing the tracking from playing on saved click
- Extended track and album titles from pushing the icons out of bounds (this happens on main)

SO9010:
Just a thought but in the ui clean-up branch we could also include some extra icons for visulizing some keyboard bindings, also we should say what they are in the settings to make people aware to how to use them. For example for the search in playlist I think we could have a little icon. We could also have a temporary refresh button introduced.

image

jacksongoode:
I like that idea, I've tried my best to figure out how overlapping triggers in druid can be prevented but I have no idea. I think this can be a PR replacing the heart with the plus. What do you think?

SO9010:
Yeah that sounds good!! I like the fact that the save button will stay larger.

PR #499, opened 2024-07-22
Remove track from playlist by index

jacksongoode:
Resolves https://github.com/jpochyla/psst/issues/477

### DIFF
--- a/psst-gui/src/data/playlist.rs
+++ b/psst-gui/src/data/playlist.rs
@@ -20,7 +20,7 @@ pub struct PlaylistAddTrack {
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
 pub struct PlaylistRemoveTrack {
     pub link: PlaylistLink,
-    pub position: usize,
+    pub index: usize,
 }
 
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
--- a/psst-gui/src/data/playlist.rs
+++ b/psst-gui/src/data/playlist.rs
@@ -20,7 +20,7 @@ pub struct PlaylistAddTrack {
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
 pub struct PlaylistRemoveTrack {
     pub link: PlaylistLink,
-    pub track_id: TrackId,
+    pub position: usize,
 }
 
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
--- a/psst-gui/src/data/playlist.rs
+++ b/psst-gui/src/data/playlist.rs
@@ -20,7 +20,7 @@ pub struct PlaylistAddTrack {
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
 pub struct PlaylistRemoveTrack {
     pub link: PlaylistLink,
-    pub index: usize,
+    pub track_pos: usize,
 }
 
 #[derive(Clone, Debug, Data, Lens, Deserialize)]
--- a/psst-gui/src/data/track.rs
+++ b/psst-gui/src/data/track.rs
@@ -25,6 +25,8 @@ pub struct Track {
     pub local_path: Option<Arc<str>>,
     pub is_playable: Option<bool>,
     pub popularity: Option<u32>,
+    #[serde(skip)]
+    pub added_index: usize,
 }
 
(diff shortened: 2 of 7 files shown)


jacksongoode:
@SO9010 Have a look at this, if it seems to work for you we can merge it.

SO9010:
It doesnt seem to work on my end! Also we really need to figure out a way to fix the notifications, because it covers the screen after a bit ahaha!

Here is a screen recording of it:


https://github.com/user-attachments/assets/42ad4227-b0cb-4a61-b2c5-b27e5bdc12d0

In fact observing it more closely it removes the song from the end of the playlist!

jacksongoode:
Oh weird, maybe I messed up. Maybe the index is wrong here, I'm sure this method will work but maybe I'll need to debug.

jacksongoode:
Opps, I think it's working now, misread the API.

SO9010:
Sadly, it still removes all the songs with the same name!

https://github.com/user-attachments/assets/0cf70406-a892-47c8-b379-51d8be8c364e

Sadly I think this is an API issue, perhaps instead we could have a check if someone is adding a song to a playlist, whether that song is already in the playlist then ask the user if they still want to add it a bit like what the official client does? 

It seems like we aren't the only one with this issue
https://community.spotify.com/t5/Spotify-for-Developers/How-delete-one-or-more-elements-from-playlist/td-p/5185630

SO9010:
Or we could do a complex thing where we long the amount of copies there are and their position and just readd them after?

kingosticks:
> Sadly, it still removes all the songs with the same name!
> 2024-07-23.11-46-41.mp4
> 
> Sadly I think this is an API issue, perhaps instead we could have a check if someone is adding a song to a playlist, whether that song is already in the playlist then ask the user if they still want to add it a bit like what the official client does?
> 
> It seems like we aren't the only one with this issue https://community.spotify.com/t5/Spotify-for-Developers/How-delete-one-or-more-elements-from-playlist/td-p/5185630

I think https://community.spotify.com/t5/Spotify-for-Developers/Positions-field-in-JSON-body-is-ignored-when-removing-tracks/td-p/6044424 is more relevent since they recently (silently) broke the API - classic Spotify. Reading that latest thread I think jperez is saying you can still remove by ID but only if you do *not* also specify the track ID. Which is pretty fragile. His de-duplicator app is doing it an entirely different way now.

jacksongoode:
@kingosticks Right, well I think the commit I referenced was made by @jperez on May 8th while that comment was made May 2nd . I would have expected the commit before last to work given that it works for the https://spotify-dedup.com/ project in production... 

https://buymeacoffee.com/jmp/dedup-spotify-breaking-changes-api Makes me think position alone should work - but maybe it needs to be `tracks: positions: []`

kingosticks:
> @kingosticks Right, well I think the commit I referenced was made by @JMPerez on May 8th while that comment was made May 2nd . I would have expected the commit before last to work given that it works for the https://spotify-dedup.com/ project in production...
> 
> https://buymeacoffee.com/jmp/dedup-spotify-breaking-changes-api Makes me think position alone should work - but maybe it needs to be `tracks: positions: []`

Maybe, sorry, I didn't check. But remember: what works today will be silently broken tomorrow. The Spotify API is run by cowboys.

jacksongoode:
Right, I initially thought only positions need to be supplied but so far I've tried:

- `tracks: {uid, positions}`
- `tracks: {uid}, positions`
- `tracks: {positions}`
- `positions`

But done really provide the right removal.

jacksongoode:
I just undid the last commit, as I think the one before that is actually working? At least this seems like the correct behavior?

https://github.com/user-attachments/assets/a95938f8-dba8-44bd-b2d2-a766cf659eca

SO9010:
Yeah I suspect the same thing, is there a way to get the original sort of it? Because isn't the sorting done in app?

kingosticks:
I tried it in the web API console, this example request body works for me.
```
{
   "positions": [0]
} 
```

jacksongoode:
With the last commit we get the tracks original index and carry it with the playlist track - all looks good to me!

SO9010:
All looks good to me!

Issue #500, opened 2024-07-23
Add to queue alerts not disappearing


jholdnak:
**Describe the bug**
Queue alerts do not disappear when adding songs in succession.

**To Reproduce**
Add songs to queue simultaneously.

**Expected behavior**
Messages should disappear in a timely manner.

**Screenshots**
![Screenshot 2024-07-22 195819](https://github.com/user-attachments/assets/436eeea0-57a0-4d0c-8fa4-d426d0fb32f1)

**Environment**
 - OS: Windows 11 Home
 - Version: 22H2

SO9010:
Heya, this is already an issue:  #255; I've had a go looking at it. It has something to do with adding items too quickly, then the timer needs clarification and perhaps overrides itself. In #478, I partially solved this by adding a close button, but if you want to work on it more, please let me know, and I will happily help.

jacksongoode:
https://github.com/jpochyla/psst/pull/503 I think this will resolve it!

PR #503, opened 2024-07-24
Fix alert timer by adding controller

jacksongoode:
Should resolve https://github.com/jpochyla/psst/issues/500


SO9010:
Just tested, works perfectly for me :P

PR #504, opened 2024-07-26
Make volume slider one line

jacksongoode:
It never made any sense to me why the volume controls were on two separate lines. It only contributed to the height of that very minimal controls panel. I think this makes a lot more sense and we don't need the volume title given that this is intuitive.

image

### DIFF
--- a/psst-gui/src/ui/mod.rs
+++ b/psst-gui/src/ui/mod.rs
@@ -132,7 +132,7 @@ fn root_widget() -> impl Widget<AppState> {
         .with_default_spacer()
         .with_child(user::user_widget())
         .center()
-        .fix_height(100.0)
+        .fix_height(88.0)
         .background(Border::Top.with_color(theme::GREY_500));
 
     let sidebar = Flex::column()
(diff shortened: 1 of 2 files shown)


jacksongoode:
image

I was messing a bit with padding and actually this makes me feel even cozier? Thoughts?

jacksongoode:
@SO9010 Since this might affect your queue panel branch, it would be great if you had a look. I think reducing the two line volume to one will make reducing the height of that panel feel less cramped.

SO9010:
Just tested! It works well in my queue panel branch! However, since it now has a slight highlighting,  it would make more sense to have it so that when you click on your name, it brings up the user profile page. 

Or we could have the whole thing as the settings button, as that makes more sense; at least, to me, it does!

jacksongoode:
> Or we could have the whole thing as the settings button

Right, I thought of that, I think its a bit weird that it's such a huge button with a lot of space in between at wide windows. I'm not so sure.

jacksongoode:
@SO9010 Just going to go with the prior way the naming and settings looked with a little less padding and a bit of space between the "Connected" and the username (before it was a little oddly smushed I felt) and we have more height to spare.

I also lowered the height of the that pref box to 88 so it's even like in your PR.

PR #511, opened 2024-08-08
Add buttons and functionality

SO9010:
Adds two buttons to help usability. 

image


psst-gui/src/widget/icons.rs
```diff
@@ -94,6 +94,13 @@ pub static PLAY_LOOP_ALL: SvgIcon = SvgIcon {
     op: PaintOp::Fill,
 };
 
+// Font awesome - reload
```
jacksongoode, line 97: We should get all icons from SF Pro Regular https://developer.apple.com/sf-symbols/

SO9010: If you want, on Saturday I can go through and replace all of the icons with those ones?

SO9010: > If you want, I can go through and replace all of the icons with those ones?

Hold on, I am unable to access any of those symbols as I don't have an Apple machine!

jacksongoode: https://github.com/brendanballon/sfsymbols-svg/tree/master/symbols

I think the SVG's can be found at other places too!

SO9010: I think I will rewrite the way that the program processes the SVG files so then we are able to place all of the files into a directory rather than having to extract the curve. It would save a lot of time in the end!

jacksongoode: Hmm, I think part of the reason was that the SVG paths alone took up less size and might be quicker to load?

SO9010: Ok, do you know if there's an efficient way to move the "view" of the path as for example with the gear icon it doesn't start at the top left meaning that it's strongly displaced.

SO9010: @jacksongoode how does this look?

jacksongoode: These icons still look quite big from a UI consistency pov. I also think they need a little padding so they don't hit the right side too hard.

PR #513, opened 2024-08-09
Truncate playable widget

jacksongoode:
I'm hoping to solve that issue where long titles and album details are not clipped.

- Refactor the track widget layout for better maintainability and ability to truncate the top and bottom row
- Fix the "now playing" dot highhighting in the album view
- Make the "now playing" marker a little lighter when active

### DIFF
--- a/psst-core/src/oauth.rs
+++ b/psst-core/src/oauth.rs
@@ -45,8 +45,7 @@ pub fn listen_for_callback_parameter(
             handle_callback_connection(&mut stream, &tx, parameter_name);
         } else {
             log::error!("Failed to accept connection on callback listener");
-            let _ = tx.send(Err(Error::IoError(std::io::Error::new(
-                std::io::ErrorKind::Other,
+            let _ = tx.send(Err(Error::IoError(std::io::Error::other(
                 "Failed to accept connection",
             ))));
         }
@@ -97,8 +96,7 @@ fn handle_callback_connection(
         }
     } else {
         log::error!("Failed to read request line from callback.");
-        let _ = tx.send(Err(Error::IoError(std::io::Error::new(
-            std::io::ErrorKind::Other,
+        let _ = tx.send(Err(Error::IoError(std::io::Error::other(
             "Failed to read request line",
         ))));
     }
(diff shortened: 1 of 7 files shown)


jacksongoode:
@SO9010 If you're bored and want to look at something new, I've been trying to figure our how to get rid of the longstanding bug where long titles and album details don't get clipped. This is as far as I got but I feel like it's really close.

SO9010:
I worked it out, the original child (the title child) needed to not be in a flex child as that will then share out the space evenly.

SO9010:
Wait, false alarm! this then makes it so it cant clip, how weird!

jacksongoode:
Darn! Well feel free to submit commits to this branch if you figure something out!

SO9010:
I have had a longer look, and it seems like something with the re-implemented logic has caused the issue where it plays that item when you click save on an item. :O

jacksongoode:
Right I think it has, that can be resolved, I know what the issue is there to fix the save play, but the truncation was what I was trying to tackle now.

SO9010:
Okay, so I have been reading the documentation. The problem comes from the fact that a non-flex widget has an infinite size. We need either the text or the dots to be a flex widget. Otherwise, they share the space equally, and the dots will never disappear. 

It clips well when just the text is in a flex row, with no expanding and no dots. Perhaps there is a way to interrupt the drawing of the text and then append the dots to it? Or maybe we could create a controller that changes a sizedbox depending on screen size, but this causes issues when the time changes.

jacksongoode:
Right, I thought about painting the dots and appending them to the title, which would somewhat solve the issue, because then the title and the dots would truncate as a single item. It's a possibility, but I think it's a little tricky given that we might have to render the text as a painted object. I don't really want to make the text painted.

I think the other option would be to somehow wrap the text and dots in their own flex object. It's just that it's very confusing to me how non-flex and flex objects are able to be truncated. It seems like they would need the expand_width method? It's all very confusing to me...

SO9010:
I had a go at wrapping both the text and the dots into a flex widget but all that happened was that the time stamp stayed in one place.

jacksongoode:
Ugh I really want to figure this issue out. But the inclusion of that dotted marker_widget is really confusing since there's no way to maximize the title width and have the remaining width filled by the dots while also allowing the title to be truncatable.

jacksongoode:
Ah sorry about that, thanks for the revert, I never used podcasts 😆 so I didn't check!

PR #518, opened 2024-08-29
Fix macOS build

jacksongoode:


jacksongoode:
@SO9010 could you see if this build work for you on Windows? There were some changes to the window handling package.

Issue #519, opened 2024-09-02
Homepage and artist page revamp ideas


SO9010:
**Is your feature request related to a problem? Please describe.**
This is related to #268 and loosely  #476; however, I'm focusing on two other features I would love to implement here. 

I am proposing a new layout for the home page and adding some information to the artist page.

I created some mock-ups of how these could look; the homepage is inspired by Spotify, just simpler and involves only the apps which are available to us; the Artist page is inspired by the album page. 

**Home page**
![Updated_home_page](https://github.com/user-attachments/assets/eb212dae-189d-4c29-b25d-9168a607e81c)
_I forgot to add the artist names and descriptions for the daily mixes!!_
[Get top tracks/artists API](https://developer.spotify.com/documentation/web-api/reference/get-users-top-artists-and-tracks )

**Artist page**
![Updated_artist_page](https://github.com/user-attachments/assets/dca4f110-f443-479f-b201-d1544fe273b4)

**Challenges** 
The hardest part of this is related to #268, which is finding the correct APIs to get the wiki documentation and the social media links; also, how would you be able to handle it if there is no information for one of these artists? Or what happens if, say, the wiki page is only available in a different language?

**Additional context**
It would be nice to have feedback from various people so that it appeals to a broad audience. Additionally, I will set up a draft PR where I am working on this.

Issue #20, opened 2021-01-14
Proxy Support


SethDusek:
Hello, I have tried using psst but I am unable to connect to Spotify. I believe this is because Spotify now has even stricter region requirements. I tried using proxychains4, but it doesn't seem to work properly with this application. The main Spotify client has an option to select a SOCKS5/SOCKS4/HTTP proxy. Are there any plans to add this to psst as well?

jpochyla:
We could do this, yeah! As we're also using a TCP session to talk to Spotify, SOCKS would be needed, right?

SethDusek:
Yea I think so. I personally use my VPS + ssh SOCKS5 to get around the region restrictions

jpochyla:
I've looked into this a bit and while HTTP proxy would be possible to use as well, SOCKS5 would be easiest to implement. I'll get to it through the next week.

jpochyla:
I have a SOCKS5 support implemented for the TCP session and the CDN connections, but the Spotify WebAPI would need a small PR for `aspotify`, so it will take a couple more days, but it's close.

jpochyla:
Hah, sorry, that took longer then expected :) I've completely removed `aspotify` and we do our own thing now. If you're still interested and feel adventurous, try latest master. I haven't tested it yet, but setting `SOCKS_PROXY` and var to smth. like `socks5://john:smith@socks.google.com` should work.

jpochyla:
Closing, feel free to open in case of any issues!

aslafy-z:
This issue should be reopened since support for HTTP proxy is not present. I had a look to ureq, it seems that it won't be implemented from now on, see https://github.com/algesten/ureq/issues/62. What were the reasons behind the choice of ureq? It looks like https://github.com/seanmonstar/reqwest for example has many more features (including http_proxy support) and is better supported.

jpochyla:
`reqwest` is an async client, we need only sync interface.

Metacinnabar:
HTTP proxy support would be great!

PR #520, opened 2024-09-05
Homepage revamp

SO9010:
This aims to sort #519 ! Please message there if there's anything you would want to see implemented! here is the current rough rough rough state of this PR!

https://github.com/user-attachments/assets/87855e62-8457-4fe1-849a-2140945080ab


jacksongoode:
The only other project that does retrieve these made for you sections is Spotube. The hidden Spotify endpoints they discovered are here.

https://github.com/KRTirtho/spotube/blob/9b024120601c0d381edeab4460cb22f87149d0f8/lib/services/custom_spotify_endpoints/spotify_endpoints.dart

SO9010:
That has made me very excited, I also spotted a "what are your friends listening to" section! I'll happily work on working out the apis. Do you think the current drafts layout is effective, we could possibly keep the "your top artists and tracks" section at the bottom as I personally really like being able to use it.

jacksongoode:
Having a look at this now, and will give some feedback!

jacksongoode:
Overall I like it! Some thoughts:

- The thumbnails are a bit too large, I don't think its in keeping with the existing UI of the application
- Some of the descriptions are garbled. I don't think this is an issue with this PR but rather how we sanitize the descriptions as a whole. HTML tags are preserved.
- I think we need a little bit more padding between the vertical sections so it feels a little but more like the division in the search (album, playlist, tracks)

image

psst-gui/src/data/playback.rs
```diff
@@ -133,6 +135,7 @@ impl PlaybackOrigin {
 impl fmt::Display for PlaybackOrigin {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &self {
+            PlaybackOrigin::Home => f.write_str("Home"),
```
jacksongoode, line 138: Nice, we actually need this now that Home is a context.

.gitignore
```diff
@@ -8,3 +8,6 @@ cache
 *.iml
 rust-toolchain
 *.ico
+psst-gui/src/ui/artist.rs
+psst-gui/src/ui/search.rs
+psst-gui/src/webapi/client.rs
```
jacksongoode, line 13: Probably shouldn't be here? Make sure your local changes are actually tracked.

psst-gui/src/ui/artist.rs
```diff
@@ -83,6 +96,39 @@ fn async_related_widget() -> impl Widget<AppState> {
...
...Berryman, drummer and percussionist Will Champion, and manager Phil Harvey. They are best known for their live performances, having also impacted popular culture with their artistry, advocacy and achievements. \n The members of the band initially met at University College London, calling themselves Big Fat Noises and changing to Starfish, before settling on the current name. After releasing Safety (1998) independently, Coldplay signed with Parlophone in 1999 and wrote their debut album, Parachutes (2000). It featured breakthrough single \"Yellow\" and received a Brit Award for British Album of the Year and a Grammy Award for Best Alternative Music Album. The group's follow-up, A Rush of Blood to the Head (2002), won the same accolades. X&Y (2005) later saw the completion of what they considered a trilogy, being nominated for Best Rock Album as well. Its successor, Viva la Vida or Death and All His Friends (2008), prevailed in the category. Both albums were the best-selling of their years, topping the charts in over 30 countries. Viva la Vida's title track also became the first British act single to lead the Billboard Hot 100 and UK Singles Chart simultaneously in the 21st century."
```
jacksongoode, line 105: Guessing this was just for show at some point?

psst-gui/src/webapi/client.rs
```diff
@@ -70,29 +65,36 @@ impl WebApi {
...
-    fn post(&self, path: impl Display) -> Result<Request, Error> {
-        self.request("POST", path)
+    fn post(&self, path: impl Display, base_url: Option<&str>) -> Result<Request, Error> {
+        self.request("GET", base_url.unwrap_or("api.spotify.com"), path)
     }
 
-    fn delete(&self, path: impl Display) -> Result<Request, Error> {
-        self.request("DELETE", path)
+    fn delete(&self, path: impl Display, base_url: Option<&str>) -> Result<Request, Error> {
+        self.request("GET", base_url.unwrap_or("api.spotify.com"), path)
```
jacksongoode, line 97: Why were these changes made?

psst-gui/src/webapi/client.rs
```diff
@@ -452,44 +779,153 @@ impl WebApi {
...
+
+        let variables = json!( {
+            "uri": section_uri,
+            "timeZone": "Europe/London",
+            "sp_t": self.access_token().unwrap(),  // Assuming this returns a Result<String, Error>
+            "country": "GB",
+            "sectionItemsOffset": 0,
+            "sectionItemsLimit": 20,
+        });
+
```
jacksongoode, line 871: We should be getting this from the user.

psst-gui/src/webapi/client.rs
```diff
@@ -707,7 +1143,7 @@ impl WebApi {
 impl WebApi {
     // https://developer.spotify.com/documentation/web-api/reference/get-audio-analysis/
     pub fn _get_audio_analysis(&self, track_id: &str) -> Result<AudioAnalysis, Error> {
-        let request = self.get(format!("v1/audio-analysis/{}", track_id))?;
+        let request = self.get(format!("v1/audio-analysis/{}", track_id), None)?;
```
jacksongoode, line 1146: Not sure how this would work in Rust but can optional args be specified that take a default if they aren't included in the call? This is looking a little messy to change every single request.

psst-gui/src/ui/artist.rs
SO9010: Yes this just shows in the artist section, just a filler for now!! That was funny be the wiki section. Haven't put much thought to it yet though. Probably should be a different pr.

psst-gui/src/webapi/client.rs
```diff
@@ -213,6 +275,243 @@ impl WebApi {
...
+                    id: id.into(),
+                    name: item.content.data.name.clone().unwrap_or_default().into(),
+                    images: item.content.data.images.as_ref().map(|images| 
+                        images.items.iter().map(|img| data::utils::Image {
+                            url: img.sources.first().map(|s| s.url.clone()).unwrap_or_default().into(),
+                            width: None,
+                            height: None,
+                        }).collect()
+                    ),
+                    description: item.content.data.description.clone().unwrap_or_default().into(),
```
jacksongoode, line 492: We can use `sanitize_html` to remove links from the descriptions. I'll add it as a commit.

psst-gui/src/webapi/client.rs
SO9010: These are so I can parse a different base URL which is needed for the custom API request. There's probably a better way of doing this though

psst-gui/src/webapi/client.rs
SO9010: Yes I agree!

psst-gui/src/webapi/client.rs
SO9010: I was trying to find this out too. Just had a deeper look, function overloading needs an extra crate: https://docs.rs/overloadf/latest/overloadf/#:~:text=With%20a%20single%20attribute%20on,use%20at%20your%20own%20risk.&text=There%20are%20some%20features%20that,unsafe%20function%20overloading

jacksongoode:
I made a small commit that does the sanitization of tags and unifies the playlist and artist thumbnail heights, I made them a bit smaller to be more cohesive with the UI. I think we should probably sanitize the HTML at some earlier step getting the playlist and the widget should be drawing from the existing way we pull playlist information - I feel there is quite a bit of duplicate code as it stands.

SO9010:
The changes you made look lovely! Thank you!!! I agree I will try and reduce the code once the functionality is there :)

psst-gui/src/webapi/client.rs
SO9010: However, this library is based on some unstable features. Do you think this is ok? 

It does not seem to be very supported in rust: https://internals.rust-lang.org/t/justification-for-rust-not-supporting-function-overloading-directly/7012

jacksongoode: We shouldn't add any new crates unless they are small and absolutely necessary I think. If this is non-standard in Rust then this is fine.

jacksongoode:
Oh, I also wanted to mention, because I saw it in some comment in the code, that all users should be able to scroll horizontally by holding Shift and scrolling. At least that's how it works on Mac and Windows. So if you have your cursor over the horizontal scroll section, you should be able to press Shift and scroll and it'll scroll through.

SO9010:
Ah, perfect. Thank you; that makes it a lot easier for me! I've now added almost all of the home sections to the playlists. Should I use the UI for the search section to have horizontal sections?

jacksongoode:
I think that should probably be a separate PR so we can just change one view at a time. I think there's some expectation with a search to see a list of items.

SO9010:
Ok perfect thanks!

SO9010:
@jacksongoode, do you think you could look at the layout? Some formatting still needs to be done, but are all the sections good and relevant? I think that the jump-back in the section should be at the top! I will start going through it and make the code look better and more efficient.

SO9010:
Starting to get to a more polished state, all that is now needed (in my opinion please tell me otherwise) is to try and get the UI (especially in the jump back in section) to be more uniform!

jacksongoode:
https://github.com/user-attachments/assets/81802a5e-5fc2-4824-bbb9-e97bf4c5478c

Few UI fixes that would be nice is to have the height of every item to be the same. Currently you can see that they are different heights depending on the text description. I also think we can remove that "-" when there is no description for an album.

It would also be nice to figure out some better more robust way of preventing text from overflowing over 3 lines, that way we can be certain it won't hit 4 lines. I think number of characters is an indirect way of doing that but I still see some albums/playlists with a description of 4 lines.

jacksongoode:
image

One other small thing, not sure if this is true on WIndows/Linux but on macOS when the scrollbars appear on hover of the row they actually overlap the items. I think we should have a small padding on the top and bottom to allow for this and also there needs to be some padding under the section title I feel anyway.

psst-gui/src/ui/album.rs
```diff
@@ -138,16 +139,29 @@ pub fn album_widget() -> impl Widget<WithCtx<Arc<Album>>> {
...
+    } else {
+        Flex::row()
         .with_child(album_cover)
         .with_default_spacer()
         .with_flex_child(album_info, 1.0)
-        .padding(theme::grid(1.0))
-        .lens(Ctx::data());
+        .with_spacer(1.0)
+        .align_left()
+    };
```
jacksongoode, line 160: Was there a reason to change this on the default album_widget?

SO9010:
Perfect thank you so much for the feedback, much appreciated and I'll work on this asap!

psst-gui/src/ui/album.rs
```diff
@@ -99,36 +99,37 @@ fn rounded_cover_widget(size: f64) -> impl Widget<Arc<Album>> {
...
-            |album: &Arc<Album>, _| album.has_explicit(),
-            |selector: &bool, _, _| match selector {
-                true => icons::EXPLICIT.scale(theme::ICON_SIZE_TINY).boxed(),
-                false => Box::new(Flex::column()),
-            },
-        ));
+pub fn album_widget(horizontal: bool) -> impl Widget<WithCtx<Arc<Album>>> {
+    let album_cover = if horizontal { rounded_cover_widget(theme::grid(16.0)) } else { rounded_cover_widget(theme::grid(6.0)) };
+
+    let album_name = if horizontal { Flex::column() } else { Flex::row() }
```
jacksongoode, line 105: If you're doing a lot of these if/else horizontal it might be better to put them all into a single clause?

psst-gui/src/ui/artist.rs
```diff
@@ -82,18 +78,33 @@ fn async_related_widget() -> impl Widget<AppState> {
...
+    } else {
+        artist.with_flex_child(Label::raw()
+            .with_font(theme::UI_FONT_MEDIUM)
+            .lens(Artist::name), 1.0)
+    };
+    
+    let artist = if horizontal {
+        artist.fix_width(theme::grid(16.0)).padding_horizontal(theme::grid(1.0)).align_left()
+    } else {
+        artist.align_left()
```
jacksongoode, line 102: Rather than breaking them up into if else for every method applied, I think it would be more readable to put them in a single if/else horizontal statement.

psst-gui/src/ui/home.rs
```diff
@@ -15,22 +21,310 @@ use super::{
...
+            results.data.artists.is_empty()
+                && results.data.albums.is_empty()
+                && results.data.playlists.is_empty()
+                && results.data.shows.is_empty()
+        },
+        Label::new("No results")
+            .with_text_size(theme::TEXT_SIZE_LARGE)
+            .with_text_color(theme::PLACEHOLDER_COLOR)
+            .padding(theme::grid(6.0))
+            .center(),
```
jacksongoode, line 199: When does this happen?

psst-gui/src/webapi/client.rs
```diff
@@ -213,8 +271,280 @@ impl WebApi {
...
+                                    height: None,
+                                }).collect()
+                            )),
+                            description: {
+                                let desc = sanitize_str(&DEFAULT, item.content.data.description.as_deref().unwrap_or_default()).unwrap_or_default();
+                                // This is roughly 3 lines of description, truncated if too long
+                                if desc.chars().count() > 65 {
+                                    desc.chars().take(62).collect::<String>() + "..."
+                                } else {
+                                    desc
```
jacksongoode, line 468: There's gotta be a better way to do this, but I can't really see what it would be now.

psst-gui/src/webapi/client.rs
```diff
@@ -213,8 +271,280 @@ impl WebApi {
...
+                            name: Arc::from(item.content.data.name.clone().unwrap()),
+                            images: item.content.data.cover_art.as_ref().map_or_else(Vector::new, |images| 
+                                images.sources.iter().map(|src| data::utils::Image {
+                                    url: Arc::from(src.url.clone()),
+                                    width: None,
+                                    height: None,
+                                }).collect()
+                            ),
+                            publisher: Arc::from(item.content.data.publisher.as_ref().unwrap().name.clone()),
+                            description: "".into(),
```
jacksongoode, line 532: I think we might need to do some truncation of all the descriptions so we can be sure that nothing on the Home page gets too tall from the descriptions.

psst-gui/src/webapi/client.rs
```diff
@@ -452,75 +808,247 @@ impl WebApi {
...
+        
+    }
+
+    fn build_home_request(&self, section_uri: &str) -> (String, String) {
+        let extensions = json!({
+            "persistedQuery": {
+                "version": 1,
+                "sha256Hash": "4da53a78e4e98d4f3fa55698af5b751fe05ca3a1a4a526ff8147e8866ccfa49f"
+            }
+        });
```
jacksongoode, line 865: Where did this hash come from?

psst-gui/src/webapi/client.rs
```diff
@@ -452,75 +808,247 @@ impl WebApi {
...
+        let request = self.get("pathfinder/v1/query", Some("api-partner.spotify.com"))?
+            .query("operationName", "homeSection")
+            .query("variables", &json_query.0.to_string())
+            .query("extensions", &json_query.1.to_string());
+
+        // Extract the playlists
+        let result = self.load_and_return_home_section(request)?;
+        Ok(result)
+    }
+    */
```
jacksongoode, line 1025: This is okay to keep in, but maybe the commenting style can be made a little more consistent :)

jacksongoode (changes requested):
This is a quick first pass. I think the if/else horizontal stuff can be made more readable and added some other comments.

I've also seen this issue with the wrapping on the search results. Have a look and compare against main.

image

psst-gui/src/webapi/client.rs
SO9010: From the spotube file: https://github.com/KRTirtho/spotube/blob/9b024120601c0d381edeab4460cb22f87149d0f8/lib%2Fservices%2Fcustom_spotify_endpoints%2Fspotify_endpoints.dart

But maybe we should re-get it! This is the website referenced by spotube  https://www.apollographql.com/docs/graphos/operations/persisted-queries/

psst-gui/src/ui/album.rs
SO9010: Yes very true!

psst-gui/src/ui/album.rs
SO9010: Ah no there wasn't this was a slip up!

psst-gui/src/ui/home.rs
SO9010: AH, it doesn't! That is very true. (This was an oversight when copying the code from the search result widget!

SO9010:
Ok, I'm committing to some changes. I have tried to reduce the if-else statements! Also, I have made the original widgets match how they looked before. 

There are two things left to do; one is to ensure that they are all the same height, which shouldn't be too hard. I might even get this done today. 

The other is that I realised that the API request I did to get the user location and timezone gets run many times when the home page is loading, which means that sometimes the loading gets stuck. We could run this just once at set up (as reignonal and timezone location wont change often), then the user could change this later using a button or text fields in preferences.

psst-gui/src/webapi/client.rs
SO9010: I'll revert it directly to the one from spotube, then we can keep an eye on theirs and up date it from there!

SO9010:
I just cached the results rather than just running it once ever!

psst-gui/src/webapi/client.rs
SO9010: I think this is good for now, there is a contraint on the size of the size of the widget now so if it does end up onto the next line it doesn't really effect the view too much.

SO9010:
@jacksongoode hows that looking?

psst-gui/src/ui/album.rs
```diff
@@ -99,14 +98,15 @@ fn rounded_cover_widget(size: f64) -> impl Widget<Arc<Album>> {
...
+pub fn album_widget(horizontal: bool) -> impl Widget<WithCtx<Arc<Album>>> {
+    let (album_cover_size, album_name_layout) = if horizontal { (16.0, Flex::column()) } else { (6.0, Flex::row()) };
+    let album_cover = rounded_cover_widget(theme::grid(album_cover_size));
+    
+    let album_name = album_name_layout
         .with_child(
             Label::raw()
                 .with_font(theme::UI_FONT_MEDIUM)
-                .with_line_break_mode(LineBreaking::Clip)
+                .with_line_break_mode(LineBreaking::WordWrap)
```
jacksongoode, line 109: Is there a reason you switched to WordWrap for just artists? I do worry that there might be a super long artist name that will cause it to look ugly :)

psst-gui/src/ui/playback.rs
```diff
@@ -172,6 +172,8 @@ fn cover_widget(size: f64) -> impl Widget<NowPlaying> {
 
 fn playback_origin_icon(origin: &PlaybackOrigin) -> &'static SvgIcon {
     match origin {
+        // TODO add home widget
+        PlaybackOrigin::Home => &icons::PODCAST,
         PlaybackOrigin::Library => &icons::HEART,
```
jacksongoode, line 177: We probably need a home icon from the SF Pro icon pack for this context.

jacksongoode:
I made just a few cleanup changes. I reduced the height of the standard widget from 8.5 to 7 so that it's 3 lines of height. I also reduced the number of characters allowed so that we can be certain that we stick to three lines and don't overflow. I also added a small padding to the top and bottom of the title widgets to just give a little bit more vertical separation between the sections.

jacksongoode:
If those changes look good to you we can merge!

SO9010:
Thank you for these changes, i think that we need to give a little extra height to allow some extra space at the bottom, as that gets cut off by the scroll bar!

SO9010:
Cool, I changed it to be 8.0. If you are happy with it, go ahead and merge! Thank you. By the way, I enjoyed working on this!

SO9010:
Also, I was wondering if we should redo the saved podcasts section to include recommended episodes and shows. It should be pretty easy to do.

SO9010:
@jacksongoode Weirdly, I think the jump-back-in section has gone off Spotify now... How weird! I'll check back in the morning, but if it still isn't there (I mean on the Spotify website), I'll submit a quick patch. How annoying :/

SO9010:
Wait it is still a thing! Why isn't it loading the data properly anymore I haven't changed anything!? Could you check this on your machine @jacksongoode

jacksongoode:
Oh no I still see it on my end, can you debug the call?

SO9010:
Oh its just my laptop giving up! I cleared the cache and its all fine now :) sorry for the alarm bells!

jacksongoode:
What cache did you clear? I'm actually noticing the same issue @SO9010

SO9010:
@jacksongoode I deleted the whole psst folder for all the cache, but maybe if it's not so stable we should remove it from the home page?

SO9010:
So very weird that it only just became a problem and only for that section. I've just checked my machines again and it is fine on both windows and Linux ATM

jacksongoode:
Hmm I did remove the cache but still get the error...

SO9010:
@jacksongoode is this still an error? Its still ok on my ends!

arch-btw:
I'm also getting an error, it's looking for missing fields in the json file apparently. Screenshot:

psst

SO9010:
This is so odd that this has only just turned up as an issue! I submitted a PR to fix the track adding. But maybe we should revert the merge to main? I can start reworking it.

SO9010:
I'm thinking this has got to do with the API call request to get the users timezone and location. We should get the user to set this up at the start. Maybe there's a crate which can be used.

SO9010:
@arch-btw are you using a VPN I'm thinking that could be the issue as it currently uses the computers IP address

arch-btw:
> @arch-btw are you using a VPN I'm thinking that could be the issue as it currently uses the computers IP address

Yes, well sort of, I'm using the socks5 feature like this:

`SOCKS_PROXY=socks5://<ip address>:<port> psst-gui`

It was introduced here: https://github.com/jpochyla/psst/issues/20

Sorry, I wish I could help more.

SO9010:
@arch-btw don't worry, just out of curiosity how does the home layout look for you now after the most recent merge

arch-btw:
@SO9010 

Oops, I didn't see your message! Here is what it currently looks like.
By the way, these upcoming features were a really good idea, I look forward to trying them when they are ready.
Let me know if you need anything else :)

psst2

jacksongoode:
@SO9010 We should probably be also have the title appear contingent on the panel's network request actually loading.

SO9010:
Just realised that the songs on the top tracks section only show album not the artist.

arch-btw:
@SO9010 @jacksongoode 

Some good news. As it turns out, I was using spotify wrong. 

Yesterday I logged into the official spotify website and made a few changes. I can't recall exactly what I did but one thing that stood out to me was that I hadn't actually "liked" any albums (only songs). So I started doing that and organized my music a bit.

Fast forward a few hours later and psst started showing everything on the homepage in psst except for: "Your top tracks" and "Your top artists", everything else shows now.

So I think what happened was that there simply was no json about my data on the spotify end, because there was not enough data that I had ever given them.

It got me thinking, the bigger "problem" might be that since psst doesn't send a lot of telemetry, people who use psst exclusively like me, might run into this same issue. Because I've never used anything besides psst.

SO9010:
@arch-btw This makes a tonne of sense! Thank you for telling me, the top tracks and top artist issue is the same issue that you only use PSST!  

Me and @jacksongoode will soon start working on a new caching system, then perhaps we could try and tackle the task of adding spotify connect which will solve this issue :). 

I've already looked into it a bit and having the connection API separated from the core will work very well. The only issues I can see would be the recommendation page. But we can solve that by adding the whole page to queue in one API request.

PR #526, opened 2024-09-24
Credits support

jacksongoode:
You can also click on the artists to get to their page!

### DIFF
--- a/psst-gui/src/data/mod.rs
+++ b/psst-gui/src/data/mod.rs
@@ -61,6 +61,7 @@ pub use crate::data::{
     user::{PublicUser, UserProfile},
     utils::{Cached, Float64, Image, Page},
 };
+use crate::webapi::{TrackCredits, WebApi};
 
 pub const ALERT_DURATION: Duration = Duration::from_secs(5);
 
@@ -86,6 +87,8 @@ pub struct AppState {
     pub finder: Finder,
     pub added_queue: Vector<QueueEntry>,
     pub lyrics: Promise<Vector<TrackLines>>,
+    pub credits: Option<TrackCredits>,
+    pub webapi: Arc<WebApi>,
 }
 
 impl AppState {
@@ -109,6 +112,12 @@ impl AppState {
             queue: Vector::new(),
             volume: config.volume,
         };
+        let webapi = Arc::new(WebApi::new(
+            SessionService::empty(),
+            None, // Remove config.proxy_url.as_deref()
+            None, // Remove config.cache_dir.clone()
+            config.paginated_limit,
+        ));
         Self {
             session: SessionService::empty(),
             nav: Nav::Home,
@@ -169,6 +178,8 @@ impl AppState {
             alerts: Vector::new(),
             finder: Finder::new(),
             lyrics: Promise::Empty,
+            credits: None,
+            webapi,
         }
     }
 }
--- a/psst-gui/src/data/mod.rs
+++ b/psst-gui/src/data/mod.rs
@@ -61,7 +61,7 @@ pub use crate::data::{
     user::{PublicUser, UserProfile},
     utils::{Cached, Float64, Image, Page},
 };
-use crate::webapi::{TrackCredits, WebApi};
+use crate::ui::credits::TrackCredits;
 
 pub const ALERT_DURATION: Duration = Duration::from_secs(5);
 
@@ -88,7 +88,6 @@ pub struct AppState {
     pub added_queue: Vector<QueueEntry>,
     pub lyrics: Promise<Vector<TrackLines>>,
     pub credits: Option<TrackCredits>,
-    pub webapi: Arc<WebApi>,
 }
 
 impl AppState {
@@ -112,12 +111,6 @@ impl AppState {
             queue: Vector::new(),
             volume: config.volume,
         };
-        let webapi = Arc::new(WebApi::new(
-            SessionService::empty(),
-            None, // Remove config.proxy_url.as_deref()
-            None, // Remove config.cache_dir.clone()
-            config.paginated_limit,
-        ));
         Self {
             session: SessionService::empty(),
             nav: Nav::Home,
@@ -179,7 +172,6 @@ impl AppState {
             finder: Finder::new(),
             lyrics: Promise::Empty,
             credits: None,
-            webapi,
         }
     }
 }
--- a/psst-gui/src/ui/credits.rs
+++ b/psst-gui/src/ui/credits.rs
@@ -10,9 +10,11 @@ use druid::{
 };
 
 pub fn credits_window(track_title: &str) -> WindowDesc<AppState> {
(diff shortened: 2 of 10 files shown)


jacksongoode:
@SO9010 Wanted to get this one in but we still need to make sure that clicking credits while there is already a credits window, we don't spawn another.

SO9010:
@jacksongoode I'll have a look at this, but I also think the UI should be consistent with the preferences window if we have a pop-up open. Otherwise, it looks a bit out of place. 

Also, I think we should consider changing the order of the right-click context menu. The contexts in the top section are related to going to the track's artist and the album, so maybe we should place it above or below those two contexts.
I also feel that without doing this, people's muscle memory would get in the way of adding the song to their playlist, as that has always been at the bottom.

image
image
Maybe the copy link to track should get moved down to the lower section.
image
However, it could also be ordered as a list of priority to the user, and maybe we could flip it all the way around so that as soon as the user right-clicks on the track, they can add it to a playlist.  This could maybe improve UX? Up to you.

jacksongoode:
I would be happy to go with what you suggest here. I'm indifferent about the location, but it should be logical or follow Spotify?

SO9010:
Its looking quite good now!

jacksongoode:
There are still two bugs I'd like it figure out. One is that clicking on the credits to each track doesn't reliably make them appear. It seems sometimes the network gets hung up on fetching the credits. And it requires interacting with some other network connection and triggering some fetch to actually cause the credits to work again.

The other thing is that we should only have one credit window at a time where we set the credit window ID and then replace the title and content if that window exists.

psst-gui/src/webapi/client.rs
```diff
@@ -38,6 +39,43 @@ use crate::{
 
 use super::{cache::WebApiCache, local::LocalTrackManager};
 
+#[derive(Debug, Clone, Data, Lens, Deserialize)]
```
SO9010, line 42: Do these structures need to be public? They aren't being called more than once, are they? I think it would make more sense and follow the rest of the code more if we had these in the function that calls the API.

psst-gui/src/webapi/client.rs
```diff
@@ -38,6 +39,43 @@ use crate::{
...
+
+#[derive(Debug, Clone, Data, Lens, Deserialize)]
+pub struct RoleCredit {
+    #[serde(rename = "roleTitle")]
+    pub role_title: String,
+    pub artists: Arc<Vec<CreditArtist>>,
+}
+
+#[derive(Debug, Clone, Data, Lens, Deserialize)]
+pub struct CreditArtist {
```
SO9010, line 64: This could be in the lyrics file, which could then be imported into this file. This is what is usually done.

psst-gui/src/data/mod.rs
```diff
@@ -84,6 +85,8 @@ pub struct AppState {
     pub finder: Finder,
     pub added_queue: Vector<QueueEntry>,
     pub lyrics: Promise<Vector<TrackLines>>,
+    pub credits: Option<TrackCredits>,
+    pub webapi: Arc<WebApi>,
```
SO9010, line 89: I'm a bit confused about why we need a new webapi module.

psst-gui/src/webapi/client.rs
```diff
@@ -285,6 +323,20 @@ impl WebApi {
...
+        let request = self
+            .get(
+                format!("track-credits-view/v0/experimental/{}/credits", track_id),
+                Some("spclient.wg.spotify.com"),
+            )?
+            .set(
+                "client-token",
+                self.session.client_token().unwrap_or_default(),
+            );
+        let response = Self::with_retry(|| Ok(request.clone().call()?))?;
```
SO9010, line 336: With_retry actually only applies when error 429 is sent back, which means too many requests have been sent. Is this needed?

SO9010 (commented):
I've had a quick look just added my thoughts, I will add some changes that I think would be good.

SO9010:
I have submitted a pr to the repository for these changes.

jacksongoode:
@SO9010 Still seeing that the response is throttled? It's pretty odd, not sure if this endpoint is more heavily restricted than others so maybe we need to compensate for that?

SO9010:
It probably is restricted due to being experimental. What happens when it doesn't work? Is there maybe a way to detect it then just block it?

jacksongoode:
When it doesn't work it just stalls and the window doesn't get refreshed. It's actually really finicky... It also might be a consequence of Psst's inability to repair a broken network connection/get a new token?

jacksongoode:
Alright I think I've gotten everything in a good state now.

SO9010:
Just checked it looks great!

SO9010:
The only thing that I can think of is perhaps caching this, but I'm not sure it is needed, its quite a fast request anyway with not a lot of information. We could address this later...

SO9010:
@jacksongoode Something that we should apply to all of our serializing where there may not be any data in the field is by having     #[serde(skip_serializing_if = "Option::is_none")].

For example in the playlist one there is:    
 #[serde(skip_serializing_if = "Option::is_none")]
    pub images: Option<Vector<Image>>,

This would optimize the requests a lot!

jacksongoode:
Where would you recommend we add that? To all of the credits fields?

psst-gui/src/ui/credits.rs
```diff
@@ -0,0 +1,218 @@
...
+};
+use druid::{
+    widget::{Controller, CrossAxisAlignment, Flex, Label, List, Maybe, Painter, Scroll},
+    Cursor, Data, Env, Event, EventCtx, Lens, RenderContext, Target, UpdateCtx, Widget, WidgetExt,
+};
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Data, Lens, Deserialize)]
+pub struct TrackCredits {
+    #[serde(rename = "trackUri")]
```
SO9010, line 17: Instead using this we should use: #[serde(rename_all = "camelCase")]

psst-gui/src/ui/credits.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub track_uri: String,
+    #[serde(rename = "trackTitle")]
+    pub track_title: String,
+    #[serde(rename = "roleCredits")]
+    pub role_credits: Arc<Vec<RoleCredit>>,
+    #[serde(rename = "extendedCredits")]
+    pub extended_credits: Arc<Vec<String>>,
```
SO9010, line 24: For this we could use the don't deserd option

jacksongoode: Ok I made all the possible credits optional I believe.

psst-gui/src/ui/credits.rs
jacksongoode: Thanks!

jacksongoode:
@SO9010 How does that most recent commit look to you? If good, I'll merge.

SO9010:
Yes looks good to me!

PR #531, opened 2024-09-25
Show empty on error for home

SO9010:
Proposal partial fix for the inconsistent display of widgets from Spotify recommended endpoints. This could be an issue (guessing here) as they are changed constantly and perhaps aren't always available as endpoints?


jacksongoode (commented):
This looks good, it would be nice if there was some way to add an error is the whole group is empty, but given that these endpoints are flaky, I think this is a better alternative than now.

SO9010:
@jacksongoode this should be good now!

Issue #528, opened 2024-09-25
Feature request: Kiosk Mode on Raspberry Pi


Tilogorn:
**Is your feature request related to a problem? Please describe.**
As per my research `Psst` is currently the best (if not only) solution that provides a stable Spotify GUI on ARM-based systems. Unfortunately, Psst does not support a "kiosk" mode, which would allow many makers to create Spotify-based media centers without the hazzle of Desktop UI around the Psst window.

My specific use case is my 5 year old son who likes to listen to his audio books on Spotify. I am creating an embedded "Spotify Player" based on a battery-powered Raspberry Pi 5 in a custom housing with built-in speakers. He is able to learn to use the Psst GUI (thanks to album image support!), but he can't do anything with the system menus. And if he were to close the Psst window by mistake, he wouldn't know what to do.

**Describe the solution you'd like**
I'd like to have a `psst-gui --kiosk` flag that opens the GUI in fullscreen, as e.g. Chromium and Firefox offer it. Closing the Psst window or accessing the Desktop UI should not be possible anymore. See this [official tutorial](https://www.raspberrypi.com/tutorials/how-to-use-a-raspberry-pi-in-kiosk-mode/) to get an idea, if this is new to you.

**Describe alternatives you've considered**
Of course I stumbled upon Volumio, but their audio board compatibility lacks of newer boards (my WM8960 is not supported) and their system image is still based on Debian Buster, which is end-of-support since a while.

The second best option would be to run Spotify in a native browser in kiosk mode, but Psst is of course a lot more performant.

**Additional context**
- My current (and the most recent) setup is a Raspberry Pi 5 with the recent Raspberry Pi OS (Bookworm, Kernel 6.6.47) and Wayland
- An example of another GTK application offering kiosk mode in this exact setup would be the pre-installed Firefox
- I am willing to support the development of this feature financially. After all, it will improve the Christmas present for my son. :-)

SO9010:
Hello, I am happy to have a look at this 👍

Tilogorn:
Glad to hear! Let me know if I can do anything to support you. I'll be available for hardware testing of course. 
Feel free to [contact me](mailto:hauskauf-freiburg@gmx.net) regarding the compensation.

SO9010:
[...]

PR #533, opened 2024-09-25
Kiosk mode

SO9010:
This is a PR for issue #528. It is essentially done and can be run by adding the system command psst-gui -- --k or psst-gui -- --kiosk to run in kiosk mode.

What I have implemented is that it opens everything in full screen without window decorations. Additionally, the settings are shown directly after logging in with a button to close it, so the settings are set but cannot be changed as I removed the button to access settings.

I have it so that when this kiosk mode is enabled, the keyboard shortcuts don't work (even though they also only work on Mac), and it cannot be exited in this way.
 
I think that the keyboard shortcuts to access the desktop need to be done on the system level because I can't interrupt things like clicking the super key. Correct me if I'm wrong. If I'm not, this should be ready to be merged with the MR with @jacksongoode's approval.

### DIFF
--- a/psst-gui/src/main.rs
+++ b/psst-gui/src/main.rs
@@ -46,29 +46,28 @@ fn main() {
         paginated_limit,
     )
     .install_as_global();
-
-    let delegate;
-    let launcher;
-    if state.config.has_credentials() {
+    let (delegate, launcher) = if state.config.has_credentials() {
         // Credentials are configured, open the main window.
         let window = ui::main_window(&state.config);
-        delegate = Delegate::with_main(window.id);
-        launcher = AppLauncher::with_window(window).configure_env(ui::theme::setup);
+        let delegate = Delegate::with_main(window.id);
 
         // Load user's local tracks for the WebApi.
         WebApi::global().load_local_tracks(state.config.username().unwrap());
+
+        (delegate, AppLauncher::with_window(window))
     } else {
-        // No configured credentials, open the account setup.
-        let window = ui::account_setup_window();
-
-        if state.config.kiosk_mode {
-            let window = ui::kiosk_setup_window();
+        // No configured credentials, open the setup window.
+        let window = if state.config.kiosk_mode {
+            ui::kiosk_setup_window()
         } else {
-        }
+            ui::account_setup_window()
+        };
+        let delegate = Delegate::with_preferences(window.id);
 
-        delegate = Delegate::with_preferences(window.id);
-        launcher = AppLauncher::with_window(window).configure_env(ui::theme::setup);
+        (delegate, AppLauncher::with_window(window))
     };
+
+    let launcher = launcher.configure_env(ui::theme::setup);
 
     launcher
         .delegate(delegate)
--- a/psst-gui/src/main.rs
+++ b/psst-gui/src/main.rs
@@ -13,6 +13,7 @@ mod widget;
 use druid::AppLauncher;
 use env_logger::{Builder, Env};
 use webapi::WebApi;
+use std::env;
 
 use crate::{
     data::{AppState, Config},
@@ -33,7 +34,14 @@ fn main() {
 
     let config = Config::load().unwrap_or_default();
     let paginated_limit = config.paginated_limit;
-    let state = AppState::default_with_config(config);
+    let mut state = AppState::default_with_config(config);
+
+    let args: Vec<String> = env::args().collect();
+    state.config.kiosk_mode = false;
+    if !args.is_empty() && (args.contains(&"-k".to_string()) || args.contains(&"-kiosk".to_string())) {
+        state.config.kiosk_mode = true;
+    }
+
     WebApi::new(
         state.session.clone(),
         Config::proxy().as_deref(),
@@ -54,7 +62,12 @@ fn main() {
         WebApi::global().load_local_tracks(state.config.username().unwrap());
     } else {
         // No configured credentials, open the account setup.
-        let window = ui::account_setup_window();
+        let mut window = ui::account_setup_window();
+        
+        if state.config.kiosk_mode {
+            window = ui::kiosk_setup_window();
+        } 
+
         delegate = Delegate::with_preferences(window.id);
         launcher = AppLauncher::with_window(window).configure_env(ui::theme::setup);
     };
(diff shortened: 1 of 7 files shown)


jacksongoode:
This doesn't appear to fullscreen on macOS. I'm trying to figure out why. I just get a borderless small window.

jacksongoode:
@SO9010 I made a few tweaks so that on macOS this works.  I've also confirmed that the binary can be run as `psst-gui -k` or `psst-gui --kiosk`. Happy to merge if those changes look good on Windows and Linux for you

psst-gui/src/ui/home.rs
```diff
@@ -82,27 +81,24 @@ pub fn recommended_stations() -> impl Widget<AppState> {
...
-        ),
+            .with_child(
+                Scroll::new(Flex::row().with_child(playlist_results_widget())).align_left(),
+            ),
     )
 }
 
 pub fn uniquely_yours() -> impl Widget<AppState> {
-    Async::new(spinner_widget, uniquely_yours_results_widget, || {Empty})
+    Async::new(spinner_widget, uniquely_yours_results_widget, || Empty)
```
jacksongoode, line 101: There are a few linting fixes that happened automatically.

psst-gui/src/ui/mod.rs
```diff
@@ -42,19 +43,31 @@ pub mod user;
...
+                .window_size(work_area.size())
+                .set_position(druid::Point::new(0.0, 0.0));
+        }
     } else {
-        win = win.with_min_size((theme::grid(65.0), theme::grid(50.0)))
-
+        win = win
+            .window_size(config.window_size)
+            .with_min_size((theme::grid(65.0), theme::grid(50.0)));
     }
```
jacksongoode, line 69: @SO9010 In the kiosk mode, I'm setting the window state to maximize. Setting it such that it's always on top and then also setting the window size to the maximum window size of the monitor and putting the position at 0, 0.

The max window size was necessary for macOS, but it might not be necessary for Windows or Linux, I want to confirm with you that this still works on Windows/Linux. I also moved to set the default window size only when we aren't in kiosk mode.

psst-gui/src/ui/mod.rs
```diff
@@ -158,8 +172,8 @@ fn root_widget() -> impl Widget<AppState> {
         .with_child(sidebar_menu_widget())
         .with_default_spacer()
         .with_flex_child(playlists, 1.0)
-        .padding(if cfg!(target_os = "macos") {
-            // Accommodate the window controls on Mac.
+        .padding(if cfg!(target_os = "macos") && !config.kiosk_mode {
+            // Accommodate the window controls on macOS
             Insets::new(0.0, 24.0, 0.0, 0.0)
```
jacksongoode, line 177: Remove this top padding from window controls if we don't have them.

jacksongoode (approved):
Approved, pending your final check.

SO9010:
> This doesn't appear to fullscreen on macOS. I'm trying to figure out why. I just get a borderless small window.
>  

Heya same issue on Linux!

psst-gui/src/ui/mod.rs
SO9010: Thank you for this! We don't really want the main window to be always on top as when the preferences window that will not have priority over the window so it might or might not show! 

Also I was wandering, on Linux the login and preferences section is still small, this is due to your fix only being for the main route widget I believe.

I was thinking that due to it being kiosk mode the whole set up should also be in full screen!

psst-gui/src/ui/mod.rs
```diff
@@ -74,40 +74,17 @@ pub fn main_window(config: &Config) -> WindowDesc<AppState> {
...
     // Change this
     let win_size = (theme::grid(50.0), theme::grid(55.0));
 
-    // On Windows, the window size includes the titlebar.
-    let win_size = if cfg!(target_os = "windows") {
-        const WINDOWS_TITLEBAR_OFFSET: f64 = 56.0;
-        (win_size.0, win_size.1 + WINDOWS_TITLEBAR_OFFSET)
-    } else {
-        win_size
-    };
```
jacksongoode, line 67: Why was this removed?

psst-gui/src/main.rs
```diff
@@ -54,7 +59,12 @@ fn main() {
         WebApi::global().load_local_tracks(state.config.username().unwrap());
     } else {
         // No configured credentials, open the account setup.
-        let window = ui::account_setup_window();
+        let mut window = ui::account_setup_window();
+
+        if state.config.kiosk_mode {
+            window = ui::kiosk_setup_window();
+        }
```
jacksongoode, line 66: Should we conditionally define the window rather than making it mutable and then switching?

psst-gui/src/ui/mod.rs
```diff
@@ -56,21 +75,15 @@ pub fn main_window(config: &Config) -> WindowDesc<AppState> {
 }
 
 pub fn preferences_window() -> WindowDesc<AppState> {
+    // Change this
```
jacksongoode, line 78: Let's remove this

psst-gui/src/main.rs
SO9010: I think this is alright as the most likely one to happen would be the account set up rather than kiosk set up

SO9010:
Need to lint I'll do that tomorrow evening.

jacksongoode:
Linted, should be good!

SO9010:
So I believe there is not way to get this to actually work for the raspberry pi due to upstream limitations.

But as it still works for windows and Mac do you think we should still merge it then I could detail how to get it working on any Linux system (by using a different window manager). Maybe as suggested we should rename this to --start-maximize. We could perhaps then add other options like --no-prefrence-cog and then --no-title-bar and we mention the limitations that on some systems it will still show the title bar. 

What do you think @jacksongoode

jacksongoode:
Sure, I think that's probably a good approach. We could separate each one of these features into Flax that could be passed altogether to create the kiosk mode experience.

jacksongoode:
Or I'm also fine with merging this as it is? It doesn't have any immediate effect and if people want it to me more granular we can make another PR to expand it.

SO9010:
Yeah that works I can open another issue stating things we need to add/change about the readme

SO9010:
Closing due to not needing it imo

Issue #349, opened 2022-10-19
Feature request: displaying song lyrics


gershik:
**Describe the solution you'd like**
Lyrics have been available in all regions for a while. `librespot` seems to support them, so I believe they could be a good addition to the client. If a two-panel layout is implemented, the playback section could display them all the time; otherwise, a button in the playback area could open them full-screen, like in the official Spotify client.

cloudybyte:
There was some discussion regarding merging psst-cores api wrapper/player and librespot in #204, however that was 18 months ago and not much has moved since then.

SO9010:
Spotifys official API is: https://coda.io/@musixmatch/musixmatch-lyrics-api/api-methods-3

SO9010:
I will have a look into this. At the moment I'm thinking we could have another page which is accessed through a toggle then that page could have the lyrics in. 

We can also sync it I'm pretty sure to the timestamp of the song so maybe we can have a condensed view of the lyrics? 

Just brainstorming here. If anyone has any ideas feel free to say them.

SO9010:
This also has links for artists!!!

SO9010:
Wait, this doesn't work I dont think. Im pretty sure we will have to sign up to be a partner. Annoying!

SO9010:
https://github.com/raitonoberu/sptlrx does it!

jacksongoode:
There are some free APIs we can use to fetch the lyrics, we should compile and list and see if we can integrate.

kingosticks:
Am I missing something, why not do the same as librespot?
https://github.com/librespot-org/librespot/blob/118a9e1a516028b75f2dbf7c50ed8a4f28fe20f5/core/src/spclient.rs#L569
and 
https://github.com/librespot-org/librespot/pull/1084/commits/f72048e5e1a6f74c5476eef89a1de37fa7bb118f

jacksongoode:
Oh look at that, yeah you might be right. We should give it a shot!

SO9010:
Yes I just got it to work with a fixed variable for the song and the image, so that's the next things I have to change! It shouldnt take too long.

SO9010:
> Yes I just got it to work with a fixed variable for the song and the image, so that's the next things I have to change! It shouldnt take too long.

As in reverse engeering the Spotify API

SO9010:
This should now be closed. Due to #535

PR #535, opened 2024-09-30
Lyric support

SO9010:
This is my start-to-finish feature request #349; I believe I have done it in a similar way to LebreSpot.

Currently what I have done is implement the API request and a very rough UI for it. Many improvements are needed for the UI, but currently, there is a new button at the bottom left that redirects you to a page called lyrics when you click it.

I want to implement time-synced lyrics, which shouldn't be too hard.

I would say its about 50% complete.


jacksongoode:
This looks great, I would also think we should have a right click to lyrics too? Maybe that's too much? I'm also fine with not having the time syncing with the lyrics which I think might be difficult to render properly and cause CPU cycles?

SO9010:
I think having it just as a toggle button makes more sense?

jacksongoode:
Yeah that could be a nice alternative to having the line highlighted and listening for the playback location. We could just click and seek to the location defined by the lyric at that time. We would need to make sure that the lyric displayed is always connected with the current song being played. Otherwise you might see a lyrics screen for another song? I think that could be okay, we just need to make sure that clicking on the lines will only seek if the song being played is that same song.

psst-gui/src/ui/lyrics.rs
```diff
@@ -31,8 +32,11 @@ fn user_top_tracks_widget() -> impl Widget<AppState> {
                 .expand_width()
                 .padding(Insets::uniform_xy(theme::grid(2.0), theme::grid(0.6)))
                 .link()
+                // 19360
+                .on_left_click(|ctx, _, c, _| ctx.submit_command(cmd::SKIP_TO_POSITION.with(c.data.start_time_ms.parse::<u64>().unwrap())))
+            
         })},
-        error_widget,
+        || Label::new("No lyrics for this song!"),
```
jacksongoode, line 39: Ah good, I was going to suggest this would be a good error catch. However, is there a case where we know when there are no lyrics vs. getting a bad request?

psst-gui/src/ui/playback.rs
```diff
@@ -186,14 +186,22 @@ fn player_widget() -> impl Widget<Playback> {
     Flex::row()
         .with_child(
             small_button_widget(&icons::SKIP_BACK)
-                .on_left_click(|ctx, _, _, _| ctx.submit_command(cmd::PLAY_PREVIOUS)),
-        )
+                .on_left_click(|ctx, _, now_playing: &mut Playback, _| {
+                    ctx.submit_command(cmd::PLAY_PREVIOUS);
+                    ctx.submit_command(lyrics::SHOW_LYRICS.with(now_playing.now_playing.as_ref().unwrap().clone()))
```
SO9010, line 191: @jacksongoode could you help me with this I cant figure out how to get it to load the lyrics when its skipped to the next track, With this it only does something after you click it twice

jacksongoode: Is this play_previous correct? I haven't dug into this but I would think we would need to listen for that command while on the lyrics page and the song changes we should fetch the new songs lyrics.

jacksongoode: I fixed this in the most recent commit! @SO9010

psst-gui/src/controller/playback.rs
```diff
@@ -310,6 +315,12 @@ impl PlaybackController {
             },
         }));
     }
+
+    fn update_lyrics(&self, ctx: &mut EventCtx, data: &AppState, now_playing: &NowPlaying) {
+        if matches!(data.nav, Nav::Lyrics) {
+            ctx.submit_command(lyrics::SHOW_LYRICS.with(now_playing.clone()));
+        }
+    }
```
jacksongoode, line 323: Not sure if this is the best place for this but adding it to the PlaybackController allows us to link the lyrics with the song.

jacksongoode:
Looks close!
- Clicking on the line to select the timestamp it seems to work well
- The text seems a little bit small maybe we could try with text medium rather than text small
- It would be nice to have some title of the song along with the artist and album at the top of the page so there's at least some indication as to where the lyrics are coming from

Other than that, I think it's ready to go.

SO9010:
@jacksongoode I could do a similar thing to what we have ATM with the album view.

jacksongoode:
I added caching and better formatting with the titles so I think this is all read for review.

psst-gui/src/webapi/client.rs
```diff
@@ -821,33 +824,38 @@ impl WebApi {
...
 
-        let result: Cached<Root> = self.load_cached(request.clone(), "Lyrics", &track_id)?;
-        Ok(result.data.lyrics.lines)
+        let lyrics: Vector<TrackLines> = self.load(request)?;
+
+        // Save to in-memory and disk cache
+        self.cache.set_lyrics(track_id_arc.clone(), lyrics.clone());
+        self.cache.save_lyrics_to_disk(&track_id_arc, &lyrics);
+
+        Ok(lyrics)
```
jacksongoode, line 858: Was your method of caching saving it to disk as well? Is this excessive?

jacksongoode: Oh are you caching the request? Maybe that's a better way to do this. I generally think we need a caching overhaul where we cache all requests in a singular efficient method.

psst-gui/src/controller/playback.rs
SO9010: I think this is good!

psst-gui/src/webapi/client.rs
SO9010: Yes i was caching the request and the response. The way that the load_cached works is if its not there it creates a folder and i name it to the song id. I think what you have done is very similar to that function so it may be excessive. Or maybe we just keep it with load_cached to keep it the same as the other functions? Then we change the load_cached function? 

I was thinking with the caching we could include a timestamp for the chached item which then deletes the cache after a certain time otherwise it may take up a lot of space. It also could then be a setting in the settings page as an "Update time" section so the user can set how often they want it to refetch everything. 

I was also thinking that using lazy loading we could cache just say the first page (50) or so (perhaps the user could define this) which would mean that we could show it very quickly to the user then the rest is brought in when the user is at the bottom of the page. (Or perhaps at the start we could have a show more button?)

We would have to log two things, user position in the list and then also the offset needed, this would also then mean that we could save the place the user is on the playlist which would be a very nice addition.

SO9010: Also, you removed the Deserialization, which has made it so it doesn't work on my machine. Did it work on yours with out it?

SO9010:
Hello, I undid your changes to the client file as you removed the deserialization part, which made it impossible to get any songs. Also, I'm pretty sure the way you cached does essentially the same thing as the load_cached function. Do correct me if I'm wrong.

Other than that I think this is looking very good, I will add some of the song information and change the lyrics icon (that was just a placeholder).

psst-gui/src/webapi/client.rs
jacksongoode: Oh I think we should use this load_cached as the go to function. Sorry for complicating it, we can eventually integrate that method into every call.

jacksongoode:
I just reverted the methods that we didn't need.

One nice thing I think would be if the user clicked on the Show lyrics button, it would actually toggle the lyrics as opposed to showing them. This could be introduced here if you feel like you have a good sense of how to do it, or we can add it in another PR. It just seemed redundant to have the lyrics button do nothing if you're already on the lyrics page. It seems intuitive that it should be a toggle.

SO9010:
Ok I will have a look into this tomorrow!

SO9010:
I can't quite figure out how to do it. I need to have AppState as a lens for the button to see the history of navigation, and I can't figure out how to implement that without causing many errors or rewriting a lot of the code.

So maybe it's a good idea to open up a separate issue for it.

jacksongoode:
@SO9010 I just figured out a simple way of doing it. I'll merge once the builds work.

Issue #268, opened 2022-01-20
[WIP] Spotify artist URI to social pages, Wiki summary


jacksongoode:
Since seeing this feature on the roadmap, I thought I'd get the initial architecture in place to make sure this might be the best/most effective string of API calls to make.

Right now it takes two calls to reach social links, four to reach a Wiki summary:

1. MusicBrainz url lookup:
- Takes Spotify artist URL
- Returns MusicBrains ID

2. MusicBrainz artist social links:
- Takes MusicBrainz ID
- Returns social links (ie. Wikidata, YouTube, SoundCloud, Facebook, Discogs, official pages, etc.)

3. Wikidata to Wikipedia page:
- Takes Wikidata ID
- Returns Wikipedia pages by language

4. Fetch summary:
- Takes Wikipedia page
- Returns summary of artist page

I feel like it _should_ be possible to reach the MusicBrainz social links with one call, but their API docs are a bit of a mess. The total call time with `reqwests` is around 3-4s.

jacksongoode:
> How are you getting the musicbrainz ID from a Spotify URL? Is there some API to do that that I've missed?

The MusicBrainz ID is retrieved by looking up which artist has the Spotify URL in their metadata on MusicBrainz.

kingosticks:
Excellent, thank you!

jpochyla:
That sounds great @jacksongoode!

jacksongoode:
A proof of concept using `ureq` can be checked out [here](https://github.com/jacksongoode/spotipedia/blob/master/src/main.rs). First time working with Rust so it might take me a while to integrate :)

SO9010:
Heya, spotify does this and I am probably able to use the same API request format as I did with getting the lyrics for the songs as it's the same background api provider.

![Screenshot_20241002-131956.png](https://github.com/user-attachments/assets/5230a19b-03be-4ee6-bf7b-8093f6f7c811)

jacksongoode:
Closed with #537

PR #537, opened 2024-10-07
Artist view update

SO9010:
This is a start at #268
image

This is using the Spotify API. You can right-click to get the social links, the random number of followers, monthly listeners, and world rank (usually 0 as they are too large a number away)!

The main thing that needs to be done is formatting (get the biography wrapped and adequately limited), and then perhaps we could have a text formatting adding in the monthly listeners and followers? @jacksongoode, how should you link these and present them?

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -1,5 +1,10 @@
 use std::{
-    fmt::Display, io::{self, Read}, path::PathBuf, sync::Arc, thread, time::Duration
+    fmt::Display,
+    io::{self, Read},
+    path::PathBuf,
+    sync::Arc,
+    thread,
+    time::Duration,
 };
 
 use druid::{
@@ -23,10 +28,10 @@ use psst_core::{
 
 use crate::{
     data::{
-        self, Album, AlbumType, Artist, ArtistAlbums, ArtistLink, AudioAnalysis, Cached, Episode,
-        EpisodeId, EpisodeLink, MixedView, Nav, Page, Playlist, PublicUser, Range, Recommendations,
-        RecommendationsRequest, SearchResults, SearchTopic, Show, SpotifyUrl, Track,
-        UserProfile, ArtistInfo, Image, ArtistStats, TrackLines,
+        self, Album, AlbumType, Artist, ArtistAlbums, ArtistInfo, ArtistLink, ArtistStats,
+        AudioAnalysis, Cached, Episode, EpisodeId, EpisodeLink, Image, MixedView, Nav, Page,
+        Playlist, PublicUser, Range, Recommendations, RecommendationsRequest, SearchResults,
+        SearchTopic, Show, SpotifyUrl, Track, TrackLines, UserProfile,
     },
     error::Error,
 };
@@ -820,7 +825,7 @@ impl WebApi {
             }
         });
         let extensions_json = serde_json::to_string(&extensions);
-        
+
         let variables = json!( {
             "uri": format!("spotify:artist:{}", id),
             "locale": "",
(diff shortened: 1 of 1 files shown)


jacksongoode:
This is cool and a feature I've been wanting for a while. If we can get the biographies from Spotify, we should just put it at the top of the artist page like you have it. If there's also an image, that would be nice to have as you have in the example. I do think that we could probably keep the monthly listeners and followers underneath the bio and then restrict the bio length to four lines.

And perhaps maybe we can add the ability to click on the biography such that it will pop up and present the more expansive metadata and full biography of the artist like we would with the credits view. Since some of these biographies I know are very long.

jacksongoode:
image

I attempted to clean it up. I think it would be nice if we could structure this in a way to have three columns, the album cover, the bio, and then the stats. I'm not sure if that will be the best way to do this. We could also put the stats below, but - as expected - I'm also having trouble with the wrapping of the bio.

SO9010:
This looks great! Thank you! I had the same issue with wrapping the bio text.

Do you think we should make the artist image a bit bigger? It seems too small to me right now. I originally set it to match the size of the album cover, but since the artist image is more important, maybe it should be larger.

If we don’t adjust the cover image, we could try placing the stats below it. Otherwise, the space between the cover and the bio might look a bit unbalanced.

jacksongoode:
I think we could make it bigger a bit. And yeah I ultimately think the stats should be below it in a row perhaps?

SO9010:
@jacksongoode I found the issue to all our wrapping worries!!! For some reason druid doesn't wrap raw labels!

SO9010:
@jacksongoode, annoyingly, this hasn't fixed it for the tracks widget! I was very hopeful!

SO9010:
@jacksongoode, This looks almost ready to be merged now; it just requires your final checks!

jacksongoode:
I was having a look at this this weekend and I think it would be really nice if we could structure the artist view such that there are three columns when the window is wide and only two when the window is narrow.

We can put the artist stats under the bio when the window is more narrow. But I think having a breakpoint that allows a three column view where the stats are on the far right side when it's most wide would allow the bio to look a little bit more readable at a narrow width.

@SO9010 What do you think about this? Right now it just looks a touch messy to me at full width.

SO9010:
Yeah sounds good I'll work on it asap!

SO9010:
Ok I'll do this!

psst-gui/src/ui/artist.rs
```diff
@@ -258,7 +258,7 @@ impl<T: Data, B: Widget<T>, S: Widget<T>> Widget<T> for ArtistInfoLayout<T, B, S
 
     fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, data: &T, env: &Env) -> Size {
         let max = bc.max();
-        let wide_layout = max.width > theme::grid(60.0);
+        let wide_layout = max.width > theme::grid(60.0) + theme::GRID * 2.0;
```
jacksongoode, line 261: I tried this and you'll see that it doesn't work I think because the width of the box constraints is different. @SO9010

SO9010: Ah, I didn't realise this as I'm using a tiling manager. I have updated it, but it may be a bit brute force as I just tested values until it matched.

jacksongoode:
All looks good! Thanks @SO9010 !

SO9010:
@jacksongoode I think we need to add something to this now, when there is no description, not even the cover widget shows, this make sit feel a bit inconsistent when it shows!

jacksongoode:
Yeah I've noticed this as well. We should add a fall back!

PR #540, opened 2024-10-11
Update readme

SO9010:
This is my start to updating the readme.md.

I'm using this template: https://github.com/othneildrew/Best-README-Template/blob/main/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,79 @@
-# Psst
-
-A fast Spotify client with a native GUI written in Rust, without Electron.
-Psst is still very early in development, lacking in features, stability, and general user experience.
-It's fully cross-platform, supporting Windows, Linux, and macOS.
-Contributions are welcome!
-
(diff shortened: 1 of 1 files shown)


jacksongoode:
I could be persuaded but I don't think the current readme layout and organization is bad. Maybe we can add some of the features to it but I don't think it needs to be completely reorganized?

jacksongoode:
I think we could some badges that will indicate stars or forks or contributors, but that doesn't seem completely necessary. 

We'll definitely need updated screenshots since the application has changed. And the building and roadmap need total refactors just to make sure that we can style them a bit better. Maybe the basic project structure could also be put just within the building section.

SO9010:
Ok, cool. Thank you for the feedback, and yes, thinking about it, the current layout is better; we just need to discuss what we want on the roadmap. We probably want multiple opinions on this.

Issue #539, opened 2024-10-09
Signal SIGXCPU on Linux when I build it myself


Fruitsalad:
**Describe the bug**
After compiling the program myself, I get a SIGXCPU signal (CPU time limit exceeded) whenever I try to play a track. The program then crashes. **This did not happen when I used a prebuilt binary.**

**To Reproduce**
I presume this does not happen on your device, which makes it rather difficult to reproduce. I simply ran:
```
sudo apt-get install libssl-dev libgtk-3-dev libcairo2-dev libasound2-dev
cargo build
cargo run --bin psst-gui
```
Of course that makes the process seem a bit cleaner than it really was since at first I missed out on the `apt-get install` step. Nevertheless I don't believe I did anything particularly out of the ordinary, and after properly ensuring all packages are installed (only `libssl-dev` wasn't already installed) a `cargo clean && cargo run --bin psst-gui` yields the same results.

**Environment**
 - OS: Linux Mint 21.2 Cinnamon (with kernel 5.15.0-122-generic)

Fruitsalad:
This issue isn't very important to me by the way, but I thought it was good to make record of it.

SO9010:
Heya, I'm pretty sure that this happens as a protection for your computer not to crash, so only the program crashes. This is because you are only building the "developer" build, add --release after your command, then it will remove this issue ;)

So try: `cargo run --bin psst-gui --release`

jacksongoode:
This is still a little bit weird that this happens. I don't know exactly what the resolution here is.

SO9010:
Ah ok so I have been looking through documentation, I think the way to fix this is through checking your current CPU time limit:

`ulimit -t`

Then you can set it to unlimited though this: 

`ulimit -t unlimited`

This may be wrong I haven't tested it yet.

SO9010:
No this isn't the answer, mine is already set to unlimited... wierd!

Fruitsalad:
I tried a couple of things just now:
1. Release builds work just fine for me. No problems at all.
2. All of the following commands give "unlimited" for me:
```sh
ulimit -R
ulimit -SR
ulimit -HR
ulimit -t
ulimit -St
ulimit -Ht
```
3. I also injected the following bit of code into `psst-gui` (specifically `route_widget()`):
```rust
let realtime_limit = rlimit::getrlimit(Resource::RTTIME).unwrap();
let cputime_limit = rlimit::getrlimit(Resource::CPU).unwrap();
[...]

PR #542, opened 2024-10-16
Fix for SIGXCPU

SO9010:
This fixes the issue of the program closing when in the dev build #539 

Simple solution for this is to customised the dev build settings: 
```
[profile.dev]
opt-level = 1
debug = true
lto = false
```


jacksongoode (approved):
Looks good, wondering where you got these config from?

SO9010:
I just used the release settings and played with it from there, as I knew this wasn't an issue in release mode. It's the opt-level that affects it.

PR #545, opened 2024-10-17
Fix for lyrics without timestamp

### LINKED ISSUE #544: Some song lyircs link to start of song
https://open.spotify.com/track/1gVFAOjp9U176pxyCGAfAR
Is an example.

@SO9010 Maybe we need to check if the lyrics are actually coupled with a timestamp? Then we can take off the click action if the lyrics aren't synced with the song.

SO9010:
Fix #544 by detecting if the song lyrics time is set at 0.


jacksongoode (approved):
Great!

PR #547, opened 2024-10-21
Suggested Changes

SO9010:
Here is how I think that bits could be re-arranged to align with the code style.


jacksongoode (approved):
Fantastic, I definitely zoomed through this PR, thank you for reviewing and fixing!

PR #548, opened 2024-10-23
Fix Linux compilation failing due to infinite recursion

itsjunetime:
Right now, on current main, I can't compile `psst-gui` due to an issue with overflowing trait requirements:

```sh
error[E0275]: overflow evaluating the requirement `sized_chunks::sized_chunk::Chunk<TrackLines>: std::marker::Sync`
   --> psst-gui/src/ui/search.rs:52:6
    |
52  |     .on_command_async(
    |      ^^^^^^^^^^^^^^^^
    |
    = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`psst_gui`)
    = note: required for `std::sync::Arc<sized_chunks::sized_chunk::Chunk<TrackLines>>` to implement `Send`
note: required because it appears within the type `im::fakepool::Arc<sized_chunks::sized_chunk::Chunk<TrackLines>>`
   --> /home/june/.cargo/registry/src/index.crates.io-6f17d22bba15001f/im-15.1.0/./src/fakepool.rs:128:19
    |
128 | pub(crate) struct Arc<A>(RArc<A>);
    |                   ^^^
note: required because it appears within the type `im::nodes::rrb::Entry<TrackLines>`
   --> /home/june/.cargo/registry/src/index.crates.io-6f17d22bba15001f/im-15.1.0/./src/nodes/rrb.rs:162:6
    |
162 | enum Entry<A> {
    |      ^^^^^
note: required because it appears within the type `im::nodes::rrb::Node<TrackLines>`
   --> /home/june/.cargo/registry/src/index.crates.io-6f17d22bba15001f/im-15.1.0/./src/nodes/rrb.rs:244:19
    |
244 | pub(crate) struct Node<A> {
    |                   ^^^^
note: required because it appears within the type `sized_chunks::types::SizeOdd<im::nodes::rrb::Node<TrackLines>, ()>`
   --> /home/june/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sized-chunks-0.6.5/src/types.rs:36:12
    |
36  | pub struct SizeOdd<A, B> {
    |            ^^^^^^^
note: required because it appears within the type `sized_chunks::types::SizeEven<im::nodes::rrb::Node<TrackLines>, sized_chunks::types::SizeOdd<im::nodes::rrb::Node<TrackLines>, ()>>`
   --> /home/june/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sized-chunks-0.6.5/src/types.rs:28:12
[...]

PR #549, opened 2024-10-23
Deduplicate code and fix a few clippy complaints

itsjunetime:
While I was working on #548, I noticed a few areas where code was unnecessarily copied or did extra work/computation that it didn't need to do. Most of that was in the WebApi client struct, but there were a few other instances elsewhere as well. There were also some issues that clippy was complaining about which I took care of.

As far as I can tell, everything still works just as expected after these changes.

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -66,11 +66,10 @@ impl WebApi {
     }
 
     fn access_token(&self) -> Result<String, Error> {
-        let token = self
-            .token_provider
+        self.token_provider
             .get(&self.session)
-            .map_err(|err| Error::WebApiError(err.to_string()))?;
-        Ok(token.token)
+            .map_err(|err| Error::WebApiError(err.to_string()))
+            .map(|t| t.token)
     }
 
     fn build_request(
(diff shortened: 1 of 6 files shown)


jacksongoode:
Awesome! Thank you for your contribution - I'll have a look now!

psst-gui/src/webapi/client.rs
```diff
@@ -1047,32 +1032,36 @@ impl WebApi {
...
     }
 }
 
+// From https://github.com/KRTirtho/spotube/blob/9b024120601c0d381edeab4460cb22f87149d0f8/lib%2Fservices%2Fcustom_spotify_endpoints%2Fspotify_endpoints.dart keep and eye on this and change accordingly
+const EXTENSIONS_JSON: &str = r#"{
+    "persistedQuery": {
+        "version": 1,
+        "sha256Hash": "eb3fba2d388cf4fc4d696b1757a58584e9538a3b515ea742e9cc9465807340be"
+    }
+}"#;
```
jacksongoode, line 1063: Wondering if there's a better place to put this, but here is fine for now.

psst-gui/src/webapi/client.rs
```diff
@@ -1086,156 +1075,81 @@ impl WebApi {
...
             .query("operationName", "homeSection")
-            .query("variables", &json_query.0.to_string())
-            .query("extensions", &json_query.1.to_string());
+            .query("variables", &self.build_home_request(section_uri)?)
+            .query("extensions", EXTENSIONS_JSON);
 
         // Extract the playlists
-        let result = self.load_and_return_home_section(request)?;
+        self.load_and_return_home_section(request)
+    }
```
jacksongoode, line 1111: Nice abstraction for this.

jacksongoode (approved):
Looks great! I'll merge it in!

PR #568, opened 2024-12-13
Show album cover art

jacksongoode:
This allows users to click on the cover art of a currently playing song and see it in higher quality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
-
-[[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "adler2"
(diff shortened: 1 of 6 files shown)


jacksongoode:
@SO9010 Interested in seeing if there are other suggestions. Maybe add a download button on right click? Esc button could trigger it's closing?

SO9010:
I think having a download option would bear TOS sadly.

SO9010:
I will have a better look soon :)

jacksongoode:
> I think having a download option would bear TOS sadly.

I actually don't know if this would be a TOS breach? I can't really find any information on this?

kingosticks:
> > I think having a download option would bear TOS sadly.
> 
> I actually don't know if this would be a TOS breach? I can't really find any information on this?

I think it would be. You are not supposed to store any Spotify content, anything that comes from their API. See [section 4](https://developer.spotify.com/terms#section-iv-restrictions), sub-section 3. There's a specific mention for coverart where they explicitly say it can be *temporarily* cached.

> Local caching. Do not locally cache any Spotify Content, except as strictly necessary to enhance the performance of your SDA and its functionality, and limited to the temporary caching of:
>
>    metadata and cover art; or

Obviously apps like this probably also want to cache audio data, but technically their public API doesn't actually provide that...

jacksongoode:
Specifically, I need some testing around Windows & Linux appearance of the cover art

SO9010:
Having a look on windows now :)

SO9010:
Okay, so on windows, the window size isn't quite the same as the image, so there are borders on the side, which make it look a bit odd. Also, note the fact that the window doesn't have a name. It feels incomplete having a whole window to itself; maybe we could take this as an opportunity to have a mini-player? This could be pretty simple by just adding the controls at the bottom.

Maybe there's a way to have a pop-up that doesn't have a frame, can't move, and disappears when the focus changes. A bit like a right-click menu? 

Also, another thing that is needed is for the image to have rounded corners.

image

jacksongoode:
@SO9010 I believe this now looks proper on Windows.

SO9010:
Checking it out on linux now!

SO9010:
Looks lovely, I really like the fact that you can close it with esc! I think it would be nice if we also had that for the credits window! I submitted a MR.

jacksongoode:
Super, I merged that PR in and I'll merge this!

PR #573, opened 2024-12-24
Sanitize all playlist descriptions

jacksongoode:
This should in theory cover all descriptions from playlists and sanitize them appropriately. There were still issues with some descriptions that were in the search results, and ampersands were not properly replaced with their actual character.


jacksongoode:
@SO9010 Does this look good, I did it quite quickly but just want to make sure nothing redundant is still here.

PR #574, opened 2024-12-24
Highlight track in nav context from link & fix link resolution loop

jacksongoode:
This was listed as a to-do item and often bothered me when I wanted to quickly open a Spotify link that someone had sent me. This is probably a more critical feature than I realized it to be.

This PR also fixes the issue where in navigating to a URL determined context, pressing back to navigate backwards would end up leading to resolving the URL once more, putting you in a loop you couldn't get out of.


arch-btw:
Wow, you're on a roll. Great job!

PR #579, opened 2025-01-13
Properly update version of Debian package

Self-Perfection:
Include branch name in Debian package version

Sanitize branch name by Debian version policy:
https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

---

This scratches my itch. Before this change all Debian packages were built with version 0.1.0. That prevented simple upgrades from package manager, as it refused to replace existing installation with the same version.

With this change packages built from main branch will have version like `0.1.0+r571-0`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -157,6 +157,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


jacksongoode (approved):
Going to approve this, but I sort of figure there might be some more unified way of versioning when the app becomes stable and relatively polished.

Issue #425, opened 2023-05-28
MPRIS DBus missing field xesam:albumArtist


YodaEmbedding:
```python
from pprint import pprint
from mpris2 import Player

uri = "org.mpris.MediaPlayer2.psst"
player = Player(dbus_interface_info={"dbus_uri": uri})

pprint(player.Metadata)
```

Outputs:
```
{dbus.String('mpris:artUrl'): dbus.String('https://i.scdn.co/image/ab67616d0000b2731f829ea9c2c7ffcec1a3c857', variant_level=1),
 dbus.String('mpris:length'): dbus.Int64(152560000, variant_level=1),
 dbus.String('mpris:trackid'): dbus.ObjectPath('/', variant_level=1),
 dbus.String('xesam:album'): dbus.String('Unknown', variant_level=1),
 dbus.String('xesam:artist'): dbus.Array([dbus.String('Alice In Chains')], signature=dbus.Signature('s'), variant_level=1),
 dbus.String('xesam:title'): dbus.String('We Die Young', variant_level=1)}
```

Insprill:
Duplicate of #368.

Issue #368, opened 2023-01-19
MPRIS `xesam:album` is "Unknown" when playing from album


TheLastZombie:
**Describe the bug**
When playing any song from within an album, the MPRIS album field (`xesam:album`) returns "Unknown".

**To Reproduce**
1. Play any liked song
2. Run `playerctl metadata`. `xesam:album` shows the album's title.
3. Right-click on the song and select "Go To Album"
4. Play the song again, this time from the context of its album
5. Run `playerctl metadata` again. `xesam:album` shows "Unknown".

**Expected behavior**
I expected `xesam:album` to be the same value as if I'd played the track from my liked songs.

**Screenshots**
![](https://user-images.githubusercontent.com/8181990/213428731-7b6a6004-b301-4f1b-a184-fdbb3e78b1f6.png)
![](https://user-images.githubusercontent.com/8181990/213428747-986049bc-4db0-4a85-8d0e-e4b598ffb946.png)

**Environment**
 - OS: Ubuntu 22.10
 - Version: psst-deb 346a729

**Additional context**
N/A

Insprill:
Platform-specific metadata isn't implemented in souvlaki yet https://github.com/Sinono3/souvlaki/issues/15

jacksongoode:
I think this is a small fix... It's a bit weird though that this album metadata isn't held in the track of an album.

TheLastZombie:
Just noticed this also happens on the lyrics page. Probably the same root issue.

![image](https://github.com/user-attachments/assets/4940149c-4504-4fc5-8af9-68dc6c447fc3)

jacksongoode:
Resolved with #581

PR #581, opened 2025-01-29
Fix for missing album metadata when song played from album

jacksongoode:
This probably resolves a few issues. One that I noticed was that the album metadata was unknown when we looked at the lyrics of a song that was played from the album, but outside of the context of an album if the song is played from a playlist it does have that album title. I think this could also be related to the MPRIS issue though I'm not sure #425 #368.

### DIFF
--- a/psst-gui/src/delegate.rs
+++ b/psst-gui/src/delegate.rs
@@ -1,9 +1,13 @@
+use directories::UserDirs;
 use druid::{
     commands, AppDelegate, Application, Command, DelegateCtx, Env, Event, Handled, Target,
     WindowDesc, WindowId,
 };
+use std::fs;
+use std::io::Read;
 use threadpool::ThreadPool;
 
+use crate::ui::album::DOWNLOAD_ARTWORK;
 use crate::ui::playlist::{
     RENAME_PLAYLIST, RENAME_PLAYLIST_CONFIRM, UNFOLLOW_PLAYLIST, UNFOLLOW_PLAYLIST_CONFIRM,
 };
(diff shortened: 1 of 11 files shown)


jacksongoode:
@SO9010 If you're on Linux could you do a quick check to see if this resolved the mpris album issues?

SO9010:
Yes, doing that now :)

SO9010:
Nice! It does work, it fixes #425 and #368 too!

jacksongoode:
@SO9010 Could you also just have a look through the PR. Just a sanity check in case :)

jacksongoode:
@jpochyla Would it be possible to make Samuel eligible as a reviewer? He's been a reliable contributor and reviewer over the last couple months.

SO9010:
@jacksongoode I just had a thought: It would be nice for consistency to have it so that we can also click on all the large covers within the album/artist views.

psst-gui/src/ui/mod.rs
```diff
@@ -151,21 +155,58 @@ fn account_setup_widget() -> impl Widget<AppState> {
...
+        env: &druid::Env,
+    ) {
+        if let druid::Event::WindowConnected = event {
+            ctx.request_focus();
+            ctx.set_handled();
+        }
+
+        if let druid::Event::KeyDown(key_event) = event {
+            // Handle D key for download
+            if key_event.key == KbKey::Character('d'.into()) {
```
SO9010, line 176: We also need a right-click menu for this; otherwise, it feels like a "hidden feature".

psst-gui/src/delegate.rs
```diff
@@ -210,6 +214,28 @@ impl AppDelegate<AppState> for Delegate {
...
+
+                    if let Ok(response) = ureq::get(url).call() {
+                        if let Ok(bytes) = response
+                            .into_reader()
+                            .bytes()
+                            .collect::<Result<Vec<_>, _>>()
+                        {
+                            if fs::write(&path, bytes).is_ok() {
+                                data.info_alert("Cover saved to Downloads folder.");
+                            }
```
SO9010, line 233: We should have an error_alert if its failed to save.

SO9010:
I had a look through it and added a couple of comments. I like the download feature, but it feels very hidden; it would be nicer if we could right-click on it and then download it like that. We could also say where it downloads after it has been downloaded, as that is not user-specified.

psst-gui/src/ui/mod.rs
jacksongoode: I just wanted to tuck this in as a way of "caching" the album art and probably not publicize it :) So I think we should leave the function as it is...

SO9010: Sounds good!

PR #585, opened 2025-02-08
UI: Enhance volume slider behavior

divadiahim:
This small PR aims to fix the inconsistent slider size, which is currently affected by the text size of the percentage indicator. The variation occurs due to text kerning and changes in width when the number reaches 100, which determines the slider knob to snap from under the cursor.

- Current implementation
broken

- This implementation
working


jacksongoode:
Thank you so much, this looks great!

PR #589, opened 2025-02-21
Alleviate browser blocking oauth socket from creating

gitRaiku:
By creating the socket and then opening the browser, execution is still halted until the browser is closed, but at least the oauth sequence can happen succesfully.


jacksongoode:
Thank you for this! There was something holding up this process where it would work intermittently when I first switched it over to Oauth.

PR #601, opened 2025-04-18
Lastfm scrobbler instance is not created on startup #599

### LINKED ISSUE #599: Lastfm scrobbler instance is not created on startup
**Describe the bug**
The lastfm scrobbler instance is not loaded / created on app startup. Hence it does not scrobble.
To scrobble again, I have to manually go into the preferences page and toggle OFF then ON the "toggle scrobbling" checkbox to enable it.

**To Reproduce**
- Log into lastfm and enable scrobbling
- Restart Psst
- Does not scrobble untilrecreating the instance

**Expected behavior**
Scrobbling should work directly when starting psst without having to toggle off and on the scrobbler option.

**Environment**

- OS: macos Ventura
- Version: 13.7.5

**Additional context**
Same as #598

marcfusch:
HI, I moved the instructions to update the scrobbler instance into a function that is called on app startup and is also called if the config file changes. This now fixes #599

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -39,7 +39,41 @@ pub struct PlaybackController {
     media_controls: Option<MediaControls>,
     has_scrobbled: bool,
     scrobbler: Option<Scrobbler>,
+    startup: bool,
 }
+fn init_scrobbler_instance(
+    data: &AppState
+) -> Option<Scrobbler> {
+    if data.config.lastfm_enable {
+        if let (Some(api_key), Some(api_secret), Some(session_key)) = (
+            data.config.lastfm_api_key.as_deref(),
+            data.config.lastfm_api_secret.as_deref(),
+            data.config.lastfm_session_key.as_deref(),
+        ) {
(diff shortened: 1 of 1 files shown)


arch-btw:
Hmm, is it possible that this causes the login info to be removed? Have to login to last.fm on restart:

> psst_gui::controller::playback] Last.fm scrobbling is disabled, clearing Scrobbler instance.

marcfusch:
> Hmm, is it possible that this causes the login info to be removed? Have to login to last.fm on restart:
> 
> > psst_gui::controller::playback] Last.fm scrobbling is disabled, clearing Scrobbler instance.

Actually it should not clear anything if I am not mistaken. But it may be worth the watch because my biggest question is why does the data clear data gets written only once

jacksongoode:
Can you confirm that this PR resolves both issues?

marcfusch:
> Can you confirm that this PR resolves both issues?

Nope this only solves the scrobbler instance during startup

arch-btw:
Got it, I was just able to reproduce the bug on main, so it is indeed not related to this PR.

PR #606, opened 2025-04-30
Remember last volume state when closing/reopening application

lukesnc:


jacksongoode:
This is good, but it might make sense to add some debounce so that we aren't writing to the config every time we scroll the slider.

lukesnc:
Forgive me if I'm misunderstanding the code, but I'm pretty sure what I added would work so that the write to config would only occur when the main window was being closed, which is when we'd save the config and close all windows anyway. I added debug statements on my end just to double check and it seems like the write only occurs on app exit

Pogodaanton:
Would this save the current volume in case of a crash, too?

PR #607, opened 2025-05-05
Make dbus_name unique across instances

### LINKED ISSUE #564: Crash due to dbus name collision
**Describe the bug**
```
$ RUST_BACKTRACE=1 psst-gui
...
[2024-12-06T03:42:28Z INFO  psst_core::session::access_token] access token expired, requesting
thread 'main' panicked at /home/cassie/.cargo/registry/src/index.crates.io-6f17d22bba15001f/souvlaki-0.7.3/src/platform/mpris/zbus.rs:139:29:
called `Result::unwrap()` on an `Err` value: SendError { .. }
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread 'main' panicked at library/core/src/panicking.rs:221:5:
panic in a function that cannot unwind
thread caused non-unwinding panic. aborting.
fish: Job 1, 'RUST_BACKTRACE=1 psst-gui' terminated by signal SIGABRT (Abort)
```

**Environment**
 - desktop enviroment: KDE Plasma (wayland)
 - kernel: 6.12.1-zen1-1-zen (64-bit)
 - Version: r547.0cd9904-1 (via [AUR](https://aur.archlinux.org/packages/psst-git)) 

**Additional context**

SO9010:
Fix #564 by adding random numbers to the end of the dbus name! @ernstki if you could just double check this works for me then @jacksongoode could you merge it for me! Thanks!

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -186,7 +186,7 @@ impl PlaybackController {
         };
 
         let mut media_controls = MediaControls::new(PlatformConfig {
-            dbus_name: format!("com.jpochyla.psst{:?}.{}", time::OffsetDateTime::now_utc(), rand::random::<i32>()).as_str(),
+            dbus_name: format!("com.jpochyla.psst{:?}{}", time::OffsetDateTime::now_utc(), rand::random::<i32>()).as_str(),
             display_name: "Psst",
             hwnd,
         })?;
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -41,20 +41,15 @@ pub struct PlaybackController {
     scrobbler: Option<Scrobbler>,
     startup: bool,
 }
-fn init_scrobbler_instance(
-    data: &AppState
-) -> Option<Scrobbler> {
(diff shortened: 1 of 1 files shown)


jacksongoode:
This looks great. Thanks for figuring out the fix. However, I'm wondering if we should increment the instance of the application rather than just generating a random string. Did you have any thoughts around this or why we should go with a random string?

SO9010:
I don't think we will have access to how many instances are running! We could use the time?

jacksongoode:
Hmm, this lead me to https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus
Which indicates it wasn't proper anyways, I think it should be `com.jpochyla.psst`. Does just changing it to the proper dbus format resolve this as well?

SO9010:
Ah! Makes sense, I don't think this will fix it, as the issue is when running two instances the dbus wants a unique name.

SO9010:
I've changed it to the correct name, but with the psst part having the random string and time.

jacksongoode:
Ok I made a small change so it's smaller, just 4 chars, and should be unique regardless.

SO9010:
looking good!

PR #609, opened 2025-05-06
Fix for release version

jacksongoode:
Should test release


jacksongoode:
I am not entirely sure if this will be what we need... but I'll have to see as the release gets built.

PR #613, opened 2025-05-18
Update ureq and bump all dependencies

### LINKED ISSUE #596: Update ureq to 3.0.10
We are currently stuck at version 2.10.1. We should change to the latest version. There are strong differences in the API, I have already started trying to do this, but I'm running into some issues when it comes to the URL builder as it now uses [http](https://crates.io/crates/http). 

I'm struggling with using .query() and will be looking more into it soon! 

Look at [migration guide](https://github.com/algesten/ureq/blob/main/MIGRATE-2-to-3.md) if you are curious.

@jacksongoode do you have much experience with this "new" ureq format or at least HTTP to build the URL. I will create a PR when its in fit shape unless you want to work on it with me asap?

SO9010:
@jacksongoode could I get your test and quick look over for this? It should be all fine. I tried to keep the flow to as similar as to what it was before. Which mean that I added a struct to the bottom which helps with creating the request, much like the request builder but for this need. Without this, I found that I couldn't get it to work using the request builder as when wanting to add a query it wouldnt let us do that,

Closes #596

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "cfb"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d38f2da7a0a2c4ccf0065be06397cc26a81f4e528be095826eee9d4adbb8c60f"
+dependencies = [
+ "byteorder",
+ "fnv",
+ "uuid",
+]
+
+[[package]]
 name = "cfg-expr"
 version = "0.15.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


SO9010:
Wait, I need to do psst-core! 

Also, I'm not sure about the lastfm scrobbling support. We get issues, like bringing in old ureq dependencies, and more desperately [this](https://github.com/dmfutcher/wrapped-vec/issues) which will make the whole project decreed. I think that we should focus on making our own API for this rather than relying on a third party dependencies. I will happily help...

SO9010:
Hmmm why wont it work for windows?

SO9010:
I don't quite understand the oauth crate do you think you could bump that? It looks like a lot changed @jacksongoode

SO9010:
I have no idea why windows keeps failing?

jacksongoode:
> Edit: It would be good, to reduce footprint, to use the ureq flag [docs.rs/oauth2/latest/oauth2](https://docs.rs/oauth2/latest/oauth2/) so we dont need to add reqwest!

Right! So this would mean `oauth2 = { version = "...", default-features = false }`. I'd say if the changes there are minimal you could go ahead a try to remove the reqwest dependency there too.

I'll get a chance to look today.

SO9010:
I added the changed cargo.toml to be what is needed, so if you want to work directly on this pr i think it would make sense to do so! oauth2 sadly uses ureq 2 maybe with my new-found skill of ureq 3.0 I'll update theirs too. 🤣

jacksongoode:
> Edit 2: Nevermind I think they do have to change their internal code as well!

Hmm, I see. Maybe it would be good to merge this first and then see if we can tackle the Oauth stuff.

SO9010:
Sounds good, I'll revert the last commit and then merge it!

SO9010:
Does the ureq code look ok?

jacksongoode:
Post-merge all looked good!

PR #614, opened 2025-05-20
Fix release packaging paths

jacksongoode:


SO9010:
This is looking good! We need to fix the links in the readme too!

jacksongoode:
It's kind of hard for you to know if this will work. Could you give it maybe one more review? Maybe we should use [act](https://github.com/nektos/act) to try and test?

jacksongoode:
Let's give it a shot!

SO9010:
Looks all good!

PR #615, opened 2025-05-22
Fix artist info not showing

### LINKED ISSUE #582: Sha256hash not working
The original key being used originally gotten from [Here](https://github.com/KRTirtho/spotube/blob/master/lib/services/custom_spotify_endpoints/spotify_endpoints.dart) is no longer working. 

At least it's not working for us, you may have noticed that the artist description is no longer showing, which is because of this. 

They get it from this  [sight](https://www.apollographql.com/docs/graphos/operations/persisted-queries/), I had a look, but I don't quite understand it would be nice if @jacksongoode could have a look into it? 

https://github.com/jpochyla/psst/blob/main/psst-gui/src/webapi/client.rs#L1070

SO9010:
Closes #582

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -828,21 +828,24 @@ impl WebApi {
         }
 
         let variables = json!( {
-            "uri": format!("spotify:artist:{}", id),
             "locale": "",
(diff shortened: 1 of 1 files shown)


jacksongoode (approved):
LGTM!

PR #623, opened 2025-06-09
Update home page to v2

SO9010:
@jacksongoode can you have a quick look through this, it had to be updated to v2.

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -15,6 +15,7 @@ use druid::{
 };
 
 use itertools::Itertools;
+use log::info;
 use once_cell::sync::OnceCell;
 use parking_lot::Mutex;
 use serde::{de::DeserializeOwned, Deserialize};
@@ -300,13 +301,13 @@ impl WebApi {
 
     fn load_and_return_home_section(&self, request: &RequestBuilder) -> Result<MixedView, Error> {
         #[derive(Deserialize)]
-        pub struct Welcome {
-            data: WelcomeData,
+        pub struct Root {
+            data: Data,
         }
 
         #[derive(Deserialize)]
         #[serde(rename_all = "camelCase")]
-        pub struct WelcomeData {
+        pub struct Data {
             home_sections: HomeSections,
         }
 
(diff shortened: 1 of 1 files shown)


psst-gui/src/webapi/client.rs
```diff
@@ -1100,50 +1098,52 @@ impl WebApi {
...
+                .set_base_uri("api-partner.spotify.com")
+                .header("app-platform", "WebPlayer")
+                .header("Content-Type", "application/json");
 
         // Extract the playlists
         self.load_and_return_home_section(request)
     }
 
     pub fn get_made_for_you(&self) -> Result<MixedView, Error> {
         // 0JQ5DAUnp4wcj0bCb3wh3S -> Daily mixes
```
jacksongoode, line 1145: This comment should be changed too.

psst-gui/src/webapi/client.rs
```diff
@@ -1100,50 +1098,52 @@ impl WebApi {
...
 
-        let request = &RequestBuilder::new(
-            format!("http://{}/{}", "ip-api.com", "json"),
-            Method::Get,
-            None,
-        )
-        .query("fields", "260")
-        .header("Authorization", format!("Bearer {token}"));
-
-        let result: Cached<User> = self.load_cached(request, "User_info", "usrinfo")?;
```
jacksongoode, line 1111: Why are we no longer using Cached<User>?

psst-gui/src/webapi/client.rs
```diff
@@ -300,13 +301,13 @@ impl WebApi {
...
-        pub struct Welcome {
-            data: WelcomeData,
+        pub struct Root {
+            data: Data,
         }
 
         #[derive(Deserialize)]
         #[serde(rename_all = "camelCase")]
-        pub struct WelcomeData {
+        pub struct Data {
```
jacksongoode, line 310: Was there a reason these were renamed?

jacksongoode (changes requested):
Had a few questions.

psst-gui/src/webapi/client.rs
SO9010: No sorry! I was running into issues when trying to rewrite it to work with the v2 version of the api and i was seeing if the issue was the deserde, those are just left over artifacts.

SO9010: Reverted now

psst-gui/src/webapi/client.rs
SO9010: I think I get what you mean, changed from daily mixes to made for you

psst-gui/src/webapi/client.rs
jacksongoode: Hmm, I think we ought to cache it, otherwise there will be latency in hitting `ip-api.com` everytime we load the home section.

jacksongoode (approved):
Small a few small adjustments, feel free to merge.

PR #626, opened 2025-06-09
Truncate playable widget (podcast layout fixed)

jacksongoode:
This is a second attempt at #513


jacksongoode:
@SO9010 This is the same branch as the reverted PR but if you could, the `episode.rs` is what has changed here.

SO9010:
Perfect all good now. (left is new and right is old)

image

PR #630, opened 2025-06-12
Cleaner structure for the artist bio

jacksongoode:
This cleans up the artist bio description. It makes the height of the description consistent with the artist image (in most cases). It also removes the redundant strings on the stats number and hides any stats if they are non-existant.

image

---

image

### DIFF
--- a/psst-gui/src/ui/artist.rs
+++ b/psst-gui/src/ui/artist.rs
@@ -191,13 +191,9 @@ fn artist_info_widget() -> impl Widget<WithCtx<ArtistInfo>> {
             utils::format_number_with_commas(info.stats.followers)
         }))
         .with_default_spacer()
-        .with_child(Either::new(
-            |ctx: &WithCtx<ArtistInfo>, _| ctx.data.stats.monthly_listeners > 0,
-            stat_row("Monthly Listeners:", |info: &ArtistInfo| {
-                utils::format_number_with_commas(info.stats.monthly_listeners)
-            }),
-            Empty,
-        ))
+        .with_child(stat_row("Monthly Listeners:", |info: &ArtistInfo| {
+            utils::format_number_with_commas(info.stats.monthly_listeners)
+        }))
         .with_default_spacer()
         .with_child(Either::new(
             |ctx: &WithCtx<ArtistInfo>, _| ctx.data.stats.world_rank > 0,


SO9010:
The code looks all good to me, makes sense to serialize to a number then format it later. I'll be able to test it later!

psst-gui/src/ui/artist.rs
```diff
@@ -178,35 +178,41 @@ fn artist_info_widget() -> impl Widget<WithCtx<ArtistInfo>> {
...
+            .lens(Ctx::data()),
+    )
+    .vertical();
 
     let artist_stats = Flex::column()
-        .with_child(stat_row("Followers:", |info: &ArtistInfo| {
-            format!("{} followers", info.stats.followers)
-        }))
+        .with_child(Either::new(
+            |ctx: &WithCtx<ArtistInfo>, _| ctx.data.stats.followers > 0,
```
SO9010, line 191: I think followers should still be shown even if at 0? What do you think.

psst-gui/src/ui/artist.rs
```diff
@@ -178,35 +178,41 @@ fn artist_info_widget() -> impl Widget<WithCtx<ArtistInfo>> {
...
         .with_default_spacer()
-        .with_child(stat_row("Ranking:", |info: &ArtistInfo| {
-            if !info.stats.world_rank.starts_with("0") {
-                format!("#{} in the world", info.stats.world_rank)
-            } else {
-                "N/A".to_string()
-            }
-        }));
+        .with_child(Either::new(
+            |ctx: &WithCtx<ArtistInfo>, _| ctx.data.stats.world_rank > 0,
```
SO9010, line 207: We could even make this go away if under 100 or something?

psst-gui/src/ui/artist.rs
jacksongoode: I actually haven't found an artist to test this on yet... Maybe we could find one and see? I guess we could just keep it and only hide the world rank.

psst-gui/src/ui/artist.rs
SO9010, line 1: Just had a thought. We should make ArtistInfoLayout generic. But we can leave it how it is right now, I will change it in a future pr (currently working on adding image and description to playlists)

SO9010: One thing! We need to add whether or not the biography is empty or not!

jacksongoode: Hmm, so if content doesn't exist then what happens, we hide the whole section? Then we need some logic around how to maximize the left column stats to fill in. 

We could do this. It would need a little thought.

SO9010: Yeah I think so. Couldn't it be possible to ignore the area which would occupy the space if discription length is 0

SO9010:
@jacksongoode could we push this soon I would like to have this so I can finish #635 I think we can just tackle the removal of the description section at a different point it time?

PR #631, opened 2025-06-13
Add playlist info

SO9010:
This PR aims to add the playlist image, description and owner to the top of the playlist, much like on the top of Albums and Artists. This will be useful when users have been implemented so we can allow the user to navigate to other users by clicking on the username.

### DIFF
--- a/psst-gui/src/ui/playlist.rs
+++ b/psst-gui/src/ui/playlist.rs
@@ -436,8 +436,7 @@ fn playlist_info_widget() -> impl Widget<WithCtx<Playlist>> {
         .lens(Ctx::data())
         .clip(Size::new(size, size).to_rounded_rect(4.0));
 
-    let owner_label = Label::dynamic(|p: &Playlist, _| p.owner.display_name.as_ref().to_string())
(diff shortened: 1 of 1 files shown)


SO9010:
Hey could you check this out #630 should get merged before this one. This is the start to being able to add the ability to view users

SO9010:
Yes that makes sense, I agree. I will have time to do this in a few days!

jacksongoode:
I'll have a look tonight!

SO9010:
Done and thanks!

jacksongoode:
Would you be up for me making a few changes and we can revert if you don't think they work?

jacksongoode:
I was trying this out and realized that maybe the collaborative label isn't accurate? Its always false for my playlists, even collaborative ones? Does it work for you?

SO9010:
Oh! I don't have any collaborative ones so I couldn't check!

jacksongoode:
image

I made some changes to adapt a bit the last line will show if its public/private and if its collaborative that's added like "Public • Collaborative". @SO9010 what do you think?

jacksongoode (approved):
See comment above ^

PR #632, opened 2025-06-13
Add clear cache button

jacksongoode:
This let's a user clear the cache.

https://github.com/user-attachments/assets/e9ba09fb-5453-4cb0-beca-cf7bf263bfb7

While I made this I realize the cache dirs are a little inconsistent.

```
❯ ls
album/           audio/           images/          related-artists/ TrackLines/
artist-info/     episode/         key/             track/           User_info/
```

We should update the user info and track lines ones. Also we should maybe just clear the whole dir and make all the caches handle the folder creation. But these seem like a separate PR.


SO9010:
Ah ok makes sense

PR #635, opened 2025-06-14
Podcast UI improvements

SO9010:
This takes a lot of elements from the home page. Also intending to add detail widget at the top of the show list!


SO9010:
Struggling with getting the podcast description to work?

SO9010:
What is needed now is to make it so that the description is properly displayed, this will be easy once #630 is merged.

Other features that I actually want to implement while thinking about it:

The skip controls should turn into skip by say 10s, then the control button for shuffle or replay should go away, and be replaced by a sleep timer and and the lyrics button should be hidden.

SO9010:
This is how its looking now. I will now work on changing the playback controls!
image

image

SO9010:
Hmm I think this if good for now! @jacksongoode could you have a quick look through this pr for me!

jacksongoode:
Super cool, maybe it should align more to the style of the artist. I can see that the padding of the pdocast image needs to be more consistent with the artist image.

image

image

I think its good other than the visual padding issue. Maybe it just needs a rebase on main.

SO9010:
Ah yes makes sense, I think I just need to add padding to the image! I'll do this when I can :)

jacksongoode (approved):
I also added total episodes and cleaned up a little!

SO9010:
Lovely, thank you! Can I merge it now?

jacksongoode:
Of course, all good!

PR #637, opened 2025-06-15
Separate QueueBehavior into ShuffleBehavior and LoopBehavior

TheForgottened:
This PR aims to separate the QueueBehavior into ShuffleBehavior and LoopBehavior, so we are not limited to 4 pre-defined queue states. 

The inspiration for this is that I always listen to my playlists in shuffle with repeat on, and this app wasn't allowing me to.

To accomplish this, an extra button was added to the UI and, in the code, QueueBehavior was turned into ShuffleBehavior and LoopBehavior.

One drawback I found is that the app crashes when trying to load a config file written before these changes, but honestly I don't know what the ideal fix (both in clean solution and respecting the app philosophy) would be, so I'll leave that up to the main faces behind the project.

Made in collaboration with [@JOSEALM3IDA](https://github.com/JOSEALM3IDA).


jacksongoode:
Thanks for the contribution!

Just for reference for others, this PR adds another icon to the toolbar.

image

I have a few concerns with this, but they mostly stem from visual clutter. I think part of the ethos of this application is to be quite minimal, functional and condensed as possible (in contrast to the 1st party app). Could we perhaps make the shuffle icon right clickable that would then allow the loop function (which I believe might be less commonly used) to be toggle-able on right click?

Also at smaller widths we have a visual glitch - I just think adding another icon might not be tenable here - even the lyrics icon is maybe too much as it is now.

image

TheForgottened:
Thanks for the feedback!

I understand your concern of UI clutter, as I myself feel like minimalist but useful UIs are way better. My only concern with your right-click approach is how to give the user proper feedback about which option is currently being used. If it's hidden behind a right-click, although technically functional and visually pleasing, it's not the best UX.

One thing I could think of is having the icons stacked. That would fix the visual glitch at smaller widths, but I guess it would keep cluttering the UI. Although the UX benefits are clear imho.

image

(please ignore the glitch on the circle.plus button)

Do you still believe that going with the right-click is the way or do you think theres some merit to the stacked icons?

jacksongoode:
Just out of curiosity, how does Spotify do this? Whats the visual mechanism to describe these shuffle x loop states?

SO9010:
Perhaps then a solution to it getting too cluttered would be when the window size becomes smaller we revert back to the original collapsed icon? 

Then when it's larger we split them into two?

SO9010:
Yeah sure the one with the right click menu but also having the clicking functionality of before if that makes sense?

Issue #644, opened 2025-06-18
a11y: Low color contrast between text and background under light mode


Guanran928:
**Describe the bug**
Low color contrast in the song list while using light mode.

**To Reproduce**
Start `psst`.

**Expected behavior**
Better contrast under light mode.

**Screenshots**

![Image](https://github.com/user-attachments/assets/f1fa6885-b4d9-44b8-98b8-6e99300a8d1f)

**Environment**
- OS: Windows 11
- Version: 24H2

Guanran928:
> Would a darker gray be enough?

Yeah, I think it's significantly more readable.

Issue #267, opened 2022-01-20
[Feature request]


ceceba223:
I would wish for a theme file so you can match the appearance of psst to the rest of the system.

PR #645, opened 2025-06-18
Theme: make placeholder color darker

secondlnl:
Hack?
Moving configuration of placeholder color to each theme instead of in the main setup.
Could maybe be an IF.

Tries to fix: #644


jacksongoode:
Thank you for taking this on, I think this looks better for sure and was about to do the same!

secondlnl:
Cool, ok something like this with the variable under just setup changed to grey 300
image1-8

jacksongoode:
I think this is fine for now! We will revisit the colors again in theme support #267 @SO9010

Issue #616, opened 2025-05-22
Public user viewer section


SO9010:
**Is your feature request related to a problem? Please describe.**
We have a section for essentially everything apart from this, pretty big feature.

**Describe the solution you'd like**
We could use the same styling for this as for the artist section. And maybe we add it to the left section at the top along with the albums and podcasts. Now I mention it we should also add the ability to save and unsave albums, we could also add an artist section for the followed artists, I think this has been mentioned before.

**Describe alternatives you've considered**
We should start with an implantation without the ability to follow and unfollow.

secondlnl:
Maybe make the account name clickable, moving it to the top and displaying profile picture next to it?

secondlnl:
> user's description too.
From my limited knowledge, Spotify doesn't allow users to add a bio (sadly), so do you mean the following/followers texts?

secondlnl:
> Oh! Got ya I think for a first contribution this may be a bit complex but I will still give some pointers just in case you want to have a go.

Yea ok no sorry, but good information nonetheless

PR #646, opened 2025-06-18
Adding a user view

secondlnl:
Ehm, so I tried something. 

This is not how to do it, I feel.

Context: Trying to add a user view.

Will fix: #616

### DIFF
--- a/psst-gui/src/data/mod.rs
+++ b/psst-gui/src/data/mod.rs
@@ -82,7 +82,7 @@ pub struct AppState {
     pub album_detail: AlbumDetail,
     pub artist_detail: ArtistDetail,
     pub playlist_detail: PlaylistDetail,
-    pub user_detail: UserDetail,
+    pub public_user_detail: PublicUserDetail,
     pub show_detail: ShowDetail,
     pub library: Arc<Library>,
     pub common_ctx: Arc<CommonCtx>,
@@ -164,11 +164,18 @@ impl AppState {
                 playlist: Promise::Empty,
                 tracks: Promise::Empty,
             },
-            user_detail: UserDetail {
-                artist: Promise::Empty,
-                albums: Promise::Empty,
-                user_info: Promise::Empty,
-                top_tracks: Promise::Empty,
+            public_user_detail: PublicUserDetail {
+                uri: Promise::Empty,
+                name: Promise::Empty,
+                image_url: Promise::Empty,
+                followers_count: Promise::Empty,
+                following_count: Promise::Empty,
+                is_following: Promise::Empty,
+                recently_played_artists: Promise::Empty,
+                public_playlists: Promise::Empty,
+                total_public_playlists_count: Promise::Empty,
+                allow_follows: Promise::Empty,
+                show_follows: Promise::Empty,
             },
 
             show_detail: ShowDetail {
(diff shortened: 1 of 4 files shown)


secondlnl:
NOTE: Copy-pasted thing and made it compile

jacksongoode:
No rush, but when you get a chance, posting a screenshot of an example profile in the UI would be great!

secondlnl:
Why shouldn't i use the spotify api documented on their developer website?

SO9010:
The API on the Spotify site is only really meant for the current user, not for public users in general (do correct me if I'm wrong though!). The API that Spotify actually uses - like in their Web Player - seems to have access to public playlists and a lot more.

Right now, we're using a mix of both: the official APIs they provide and the ones used by their Web Player. For this feature specifically, I think we'll need to use the latter - the one that isn't directly provided. But yeah, like I said before, I'm more than happy to help with it :)

As for the image issue - I think it's to do with how the struct is being deserialized hence the JSON error message. You may also need to use Option in case there are no images. I can't properly check the code at the moment, but if you're just calling Vec: : new() you're not actually putting any data in, right? I'll have a better look at it soon though.

SO9010:
Hey! I had a quick look at your get_publicuser_info function - it looks like you might still be calling the endpoint for artists, which isn't quite the one we need here. Did you get a chance to use the info I sent about the correct endpoint to call? Unless I'm misunderstanding something, it seems like it's just the wrong type of request. Let me know if I've got that wrong though! :)

secondlnl:
Implementing the `images` field in the webapi initializer, made the error happen, if I comment it out and then also comment out the field and its function in user data and comment out the `cover_widget` and `public_user_widget` in user UI, it works. Sorry for the mess.

SO9010:
Yeah, you just need the option keyword or https://serde.rs/attr-skip-serializing.html#:~:text=Please%20use%20the%20skip%20attribute,skip_deserializing%20to%20skip%20deserializing%20only.

SO9010:
https://serde.rs/field-attrs.html#skip

#[serde(skip_serializing_if = "path")]

SO9010:
But then you have to populate the images struct using a different API, (perhaps the one mentioned before)

secondlnl:
So the API things, how do they work?

secondlnl:
Sorry what I meant was how do you call the api, model the json and plug in all the values correctly, cause my thing/code doesn't work. Thanks for your help!

SO9010:
Ok so I've had a look through and did a few small changes, the api now works correctly, and all you need to do (still not an easy feat but you should be able to do it as you got it to work 90percent with userProfile)! So you are very close now! Great job! If you need my help again just let me know! I added PublicUserDetail which now contains.
```Rust
pub uri: String,
pub name: String,
pub image_url: String,
pub followers_count: i64,
pub following_count: i64,
pub is_following: bool,
pub recently_played_artists: Vec<RecentlyPlayedArtist>,
pub public_playlists: Vec<PublicPlaylist>,
pub total_public_playlists_count: i64,
pub allow_follows: bool,
pub show_follows: bool,
```
Which will make it far easier now ;)

secondlnl:
Okay, will have to try next next week.

Issue #636, opened 2025-06-15
No Lyrics


Areg472:
**Describe the bug**
Whenever I click the lyrics button it always shows "no lyrics available" even on songs with billions of streams

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**Environment**

- OS:
WIndows
- Version:

**Additional context**

SO9010:
@jacksongoode could you have a look at it, I realised i didnt migrate the client properly for te get lyrics and it should look more like this:
``` Rust
let request = &RequestBuilder::new(
          format!(
              "color-lyrics/v2/track/{track_id}/image/https%3A%2F%2Fi.scdn.co%2Fimage%2F{}",
              track_id.clone().split_off(3)
          ),
          Method::Get,
          None,
      )
      .set_base_uri("spclient.wg.spotify.com")
```

But when i do that I still get an error 400. It works on the older versions of psst...

jacksongoode:
Yeah I was noticing the lyrics were missing, I wasn't sure if it was the v2 series of changes? I think in your example we still need to pass in the token.

SO9010:
I tried it with replicating everything?? There must be something missing. The token is also added with load cache in the request function.

SO9010:
Ohhhh ok, strange that it still works on the older versions of psst though?

jacksongoode:
Right actually, maybe that was just for the now playing, I'm not so sure why but maybe the ureq update broke it?

jacksongoode:
Fixed in #648

PR #648, opened 2025-06-18
Fix lyrics call in request builder

jacksongoode:
I believe this should fix the issue #636. I need to test this and also make sure we don't need the extra headers.

I also took this opportunity to change the cache folders to kebab-case.


SO9010:
Works for me! The changes look all good too!

PR #650, opened 2025-06-22
Implement pipewire output

azam:
* Initial implementation for pipewire output.
* Change `psst-gui` and `psst-cli` to explicitly pass features of `psst-core` by disabling `psst-core` default features when compiled from `psst-gui` or `psst-cli`.
  This makes it easier to set features at `psst-gui` or `psst-cli` without changing `psst-core/Cargo.toml`, since currently there is no way to switch `AudioOutput` implementation at runtime.
  This should not affect current build because default features are set at `psst-gui` and `psst-cli` and is passed down as `psst-core` feature as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,6 +87,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "annotate-snippets"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ccaf7e9dfbb6ab22c82e473cd1a8a7bd313c19a5b7e40970f3d89ef5a5c9e81e"
+dependencies = [
+ "unicode-width",
+ "yansi-term",
+]
+
+[[package]]
 name = "anstream"
 version = "0.6.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


azam:
Hi,
I also have a branch with selectable backend, buildable with `cpal` and `pipewire` but not tested for `cubeb` at https://github.com/azam/psst/tree/feature/pipewire-selectable if you want to check it out. Need a restart of the app after selection change though.
Screenshot_20250625_031153

azam:
@jacksongoode 
Would you consider pulling this in?
I can either proceed with the output selection on a separate PR.

jacksongoode:
I'm a little curious if you've checked out https://github.com/SolarLiner/interflow? It would be another backend, but it might have the possibility of replacing cpal and support more than just pipewire (also coreaudio, alsa, WASAPI). I would be hesitant to provide a path to pipewire support via its own backend just for the sake of maintenance.

azam:
@jacksongoode 
Managed to get interflow to output audio, but the sample counts are borked (channel count are not synced, so playback is 0.5x slow. Probably will try with a fixed channel count later).

I will try again later on the next weekend, in the meantime here is the branch if you want to take a look or help me with it.
 
https://github.com/azam/psst/tree/feature/interflow

azam:
@jacksongoode 
Sure, I intend to do that once I figure out the sampling issue.

PR #661, opened 2025-06-29
Add icons to sidebar menu links

Cleboost:
Updated the sidebar menu to display icons next to each link (Home, Tracks, Albums, Podcasts) by modifying the sidebar_link_widget to accept and render an optional icon.

{3A1B3B7D-F50E-4F58-AA27-4DE1B34D9E55}
{C595643F-FB48-4874-AFA5-8A1EC34E8A46}

### DIFF
--- a/psst-gui/src/ui/mod.rs
+++ b/psst-gui/src/ui/mod.rs
@@ -394,19 +394,40 @@ fn sidebar_menu_widget() -> impl Widget<AppState> {
     Flex::column()
         .with_default_spacer()
         .with_child(sidebar_link_widget("Home", Some(&icons::HOME), Nav::Home))
-        .with_child(sidebar_link_widget("Tracks", Some(&icons::MUSIC_NOTE), Nav::SavedTracks))
-        .with_child(sidebar_link_widget("Albums", Some(&icons::ALBUM), Nav::SavedAlbums))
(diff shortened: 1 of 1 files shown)


jacksongoode:
Just a passing thought, the icons on the left of the sections are a nice touch. I'm not so sure about any of the changes to the search bar... I think having more space to use for your search query would be better than an icon.

SO9010:
> If the PR is merged, I will submit a separate PR to introduce compact mode for the whole application. l

I'm a big fan of this idea! The compact mode could also have sub settings such as removing the info parts of say playlists and podcasts. Or changing the padding too? What do you think @jacksongoode

jacksongoode:
> I think it should be an option in settings to enable/disable

Hm yeah, I think we could sneak it under the show/hide covers.

As for other customization, since minimalism is the ethos of Psst (I feel). I wouldn't want to provide so much user facing customization given that it would also likely lead to breaking the interface. I think visual changes should be made sparingly and thoughtfully - though I'm not opposed to providing a stylesheet/config file that people could fiddle with all the constants at their own peril if they so desire.

jacksongoode:
Here are the icons at `16.0`:
image

And here at `18.0`:
image

While 16 keeps the same height, the icons look at bit small. I lean towards 18.

I'll give this a day or so for folks to comment on this style change and then merge.

Cleboost:
You're right, 18 is better I think

PR #664, opened 2025-06-30
Add Windows taskbar playback controls integration

### LINKED ISSUE #659: Windows Taskbar Playback Controls
The most convenient way for me to control Spotify while using the official app was through the taskbar popup because I wouldn't have to open the app entirely and navigate to the pause/play button.

I would like playback controls added to the taskbar hover popup on Windows just like the official Spotify app.

Opening the app and pressing the playback controls on there which is much less convenient.

**Additional context**
What it currently looks like on Psst:
Image
Example from the official Spotify app:
Image

Cleboost:
Introduces a `TaskbarManager` for Windows that adds playback control buttons (play/pause, next, previous) to the taskbar thumbnail toolbar. Updates playback controller to initialize and update these buttons based on playback state, and adds a new `PLAY_PAUSE_OR_RESUME` command for unified play/pause handling. Updates dependencies to include the windows crate and related packages.

close https://github.com/jpochyla/psst/issues/659

{E539EC52-B697-4595-B983-B2877D835BBE}
{13E43559-D2AA-402B-B642-4427473296A0}

### DIFF
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -255,7 +255,7 @@ impl PlaybackController {
                     if let Err(e) = taskbar_manager.update_all_buttons(playback_state) {
                         log::error!("Failed to update taskbar buttons: {:?}", e);
                     }
-                }   
+                }
             }
             self.last_taskbar_state = Some(playback_state);
         }
--- a/psst-gui/src/controller/playback.rs
+++ b/psst-gui/src/controller/playback.rs
@@ -25,6 +25,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::{
     cmd,
+    controller::taskbar::TaskbarManager,
     data::Nav,
     data::{
         AppState, Config, NowPlaying, Playable, Playback, PlaybackOrigin, PlaybackState,
@@ -38,6 +39,9 @@ pub struct PlaybackController {
     thread: Option<JoinHandle<()>>,
     output: Option<DefaultAudioOutput>,
     media_controls: Option<MediaControls>,
+    taskbar_manager: Option<TaskbarManager>,
+    taskbar_buttons_initialized: bool,
+    last_taskbar_state: Option<PlaybackState>,
     has_scrobbled: bool,
     scrobbler: Option<Scrobbler>,
     startup: bool,
@@ -75,6 +79,9 @@ impl PlaybackController {
             thread: None,
             output: None,
             media_controls: None,
+            taskbar_manager: None,
+            taskbar_buttons_initialized: false,
+            last_taskbar_state: None,
             has_scrobbled: false,
             scrobbler: None,
             startup: true,
@@ -104,6 +111,9 @@ impl PlaybackController {
             .map_err(|err| log::error!("failed to connect to media control interface: {:?}", err))
             .ok();
 
+        self.taskbar_manager = TaskbarManager::new(window, event_sink.clone(), widget_id)
+            .map_err(|err| log::error!("failed to initialize taskbar manager: {:?}", err))
+            .ok();
         self.sender = Some(player.sender());
         self.thread = Some(thread::spawn(move || {
             Self::service_events(player, event_sink, widget_id);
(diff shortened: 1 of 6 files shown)


psst-gui/src/controller/taskbar.rs
```diff
@@ -0,0 +1,477 @@
+//! Windows Taskbar Thumbnail Toolbar integration
+//!
+//! This module provides functionality to display playback control buttons (https://github.com/jpochyla/psst/issues/659)
+//! (play/pause, next, previous) in the Windows taskbar thumbnail preview.
```
jacksongoode, line 4: I was just wondering if this code was template code that was referenced from another project? It seems like something we could at least check how another project does it as well.

psst-gui/src/controller/playback.rs
```diff
@@ -447,6 +484,7 @@ where
 
                 if let Some(queued) = data.queued_entry(*item) {
                     data.start_playback(queued.item, queued.origin, progress.to_owned());
+                    self.setup_taskbar_buttons_on_first_play(PlaybackState::Playing);
```
jacksongoode, line 487: Why do we need this here and below on line 536? This section is for the queued media?

jacksongoode (commented):
This looks good. I haven't gone through it extensively, but I am curious about how the taskbar manager needs to be initialized on the first play. Is this a necessity or could we initialize it without first playing any media?

I'll review again when there's a little more time.

psst-gui/src/controller/taskbar.rs
Cleboost: I absolutely didn't take inspiration from other projects, but it's a good idea to see how others do it.

psst-gui/src/controller/playback.rs
Cleboost: I only display the buttons once the first track has started playing. This prevents having controls that don't do anything. It also ensures that the window is fully loaded before injecting the buttons.

I'm not sure it's the right place to call the function, but it worked, so I didn't think too much about it.

NarwhalKid:
For the record, this is what the controls look like on Spotify when there's no song playing. (Clicking the play/pause button swaps the display button between play and pause but doesn't do anything playback wise)
image

jacksongoode (approved):

jacksongoode:
Ok I did test and it works! I'd still like to review the code since it's quite a lot. @Cleboost did you use a reference for this taskbar code?

Cleboost:
Some Reddit subs and docs

psst-gui/src/controller/playback.rs
```diff
@@ -602,6 +647,17 @@ where
...
+                let event_sink = ctx.get_external_handle();
+                let widget_id = ctx.widget_id();
+                std::thread::spawn(move || {
+                    std::thread::sleep(std::time::Duration::from_millis(100));
+                    let _ = event_sink.submit_command(
+                        cmd::INITIALIZE_TASKBAR,
+                        (),
+                        Target::Widget(widget_id),
+                    );
+                });
```
jacksongoode, line 659: Why do we need to wait here?

psst-gui/src/controller/taskbar.rs
```diff
@@ -0,0 +1,516 @@
...
+    Win32::UI::WindowsAndMessaging::*,
+};
+
+#[cfg(windows)]
+use raw_window_handle::{HasRawWindowHandle, RawWindowHandle, Win32WindowHandle};
+
+#[cfg(windows)]
+use crate::{cmd, data::PlaybackState};
+
+#[cfg(windows)]
```
jacksongoode, line 22: I think you could make this whole file windows only for now rather than each of these decorators.

psst-gui/src/controller/taskbar.rs
```diff
@@ -0,0 +1,516 @@
...
+                if opaque {
+                    *p = 0xFFFFFFFF;
+                } else {
+                    *p = 0x00000000;
+                }
+            }
+        }
+    }
+
+    unsafe fn icon_from_polys(polys: &[&[(i32, i32)]]) -> HICON {
```
jacksongoode, line 476: Hmm, I'm specifically curious about your source for these methods to make the media buttons?

jacksongoode (commented):
Few thoughts, but some references in links would help me refer a bunch!

psst-gui/src/controller/taskbar.rs
Cleboost: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createiconindirect
https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-iconinfo

It was my first time developing a taskbar controller. I read a lot of docs and did a lot of experimenting.

psst-gui/src/controller/playback.rs
Cleboost: I think I put that in because we had to wait for something, otherwise things would have gone badly. I don't remember exactly.

PR #666, opened 2025-06-30
Add play button to album and playlist views

### LINKED ISSUE #656: Add the Play button
Basically whatever this button does on any albums/artists/playlists page
Image

Cleboost:
Introduces a play button to both album and playlist detail views, allowing users to start playback directly from these screens. Implements new commands and delegate handling to support playing all tracks from an album or playlist, respecting the current queue behavior (e.g., random or sequential).

> Not sure about code quality or used practice

close https://github.com/jpochyla/psst/issues/656

{962628E1-A986-46AE-A412-615344ABF36F}
{5470D146-7C18-4D6E-8B40-FE0B4A3F4F56}

### DIFF
--- a/psst-gui/src/delegate.rs
+++ b/psst-gui/src/delegate.rs
@@ -3,7 +3,9 @@ use druid::{
     commands, AppDelegate, Application, Command, DelegateCtx, Env, Event, Handled, Target,
     WindowDesc, WindowId,
 };
+use rand::seq::IndexedRandom;
 use std::fs;
+use std::sync::Arc;
 use threadpool::ThreadPool;
 
 use crate::ui::playlist::{
@@ -13,7 +15,7 @@ use crate::ui::theme;
 use crate::ui::DOWNLOAD_ARTWORK;
 use crate::{
     cmd,
-    data::{AppState, Config},
+    data::{AppState, Config, Track},
     ui,
     webapi::WebApi,
     widget::remote_image,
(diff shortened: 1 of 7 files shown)


Cleboost:
Can't dev for artist, I can't access to view
https://github.com/user-attachments/assets/a7bfe789-80df-4664-8ea2-904a4873c152

jacksongoode (approved):
This looks great, I'll test shortly

jacksongoode:
It would be great to have a different play icon for this as well instead of the duplicate.

jacksongoode:
@SO9010 Could you have a look over this code, if there are any issues in the implementation?

psst-gui/src/delegate.rs
```diff
@@ -325,3 +352,34 @@ impl Delegate {
...
+    origin: crate::data::PlaybackOrigin,
+    queue_behavior: crate::data::QueueBehavior,
+    to_playable: F,
+) where
+    F: Fn(&T) -> crate::data::Playable,
+{
+    if !items.is_empty() {
+        let playables: Vec<_> = items.iter().map(&to_playable).collect();
+        let is_random = queue_behavior == crate::data::QueueBehavior::Random;
+        let position = if is_random {
```
SO9010, line 368: Rather than define is_random why not just use if queue_behavior == crate::data::QueueBehavior::Random {
..
}

psst-gui/src/delegate.rs
```diff
@@ -325,3 +352,34 @@ impl Delegate {
...
+fn play_items_with_mode<T, F>(
+    ctx: &mut DelegateCtx,
+    items: &[T],
+    origin: crate::data::PlaybackOrigin,
+    queue_behavior: crate::data::QueueBehavior,
+    to_playable: F,
+) where
+    F: Fn(&T) -> crate::data::Playable,
+{
+    if !items.is_empty() {
```
SO9010, line 365: Could be neater if you just had if items.is_empty() {
Return;
}

SO9010 (dismissed):

jacksongoode:
> we could perhaps add the lyrics button to the top

The bottom playback bar is unique to the currently playing track, we want the lyrics button to be anchored to that. The play button for a view like a playlist operates similarly in theory to the sort (which sorts the viewed list). So I don't think the lyrics button should be moved.

SO9010:
> Right now I think sorting might be broken :( 
Just submitted a PR to fix it :P

SO9010:
Yeah I'm a fan of greying it out or even make it disappear when not needed. I think that would be good to have with the lyrics button too. For example its not needed in podcasts.

Cleboost:
Any news for this features?

SO9010:
Personally, I don't like it staying as a play button, if we want it to restart the track, I think we have to make it so that at least the icon changes to represent the restarting of the track. I also still feel like the button is too small. @jacksongoode what do you think as we have different ideas on how this should be apporached.

PR #675, opened 2025-07-14
Fix refreshing page and make refresh to ctrl r

### LINKED ISSUE #674: Broken playlist sorting
As the title suggests, the playlist sorting option doesn't re-organise the playlist items anymore!

Edit: cmd::NAVIGATE_REFRESH seems to be broken

SO9010:
Fixes #674 @jacksongoode can you have a quick skim through, I made two functions, one for refreshing all the pages universally and then one just for playlists for the playlist sorting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,7 @@ jobs:
             target: x86_64-unknown-linux-gnu
           - os: ubuntu-latest
             target: aarch64-unknown-linux-gnu
-          - os: macOS-latest
+          - os: macos-latest
           - os: windows-latest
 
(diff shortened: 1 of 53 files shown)


jacksongoode:
Ah so does the sort function now call the refresh?

SO9010:
Thats how it was done before I just stopped working for some reason so I just reimplemented it in a different way.

SO9010:
But I do agree it should be done differently. We just wan tto refresh the data on the page to reflect the sorting.

SO9010:
@jacksongoode do you think we could merge this for now? And we open an issue/discussion, on how we should tackle this in the future?

PR #676, opened 2025-07-14
Update client to force sending the body allowing use of the updated api

### LINKED ISSUE #669: Saving/removing track to/from library is broken
**Describe the bug**
Since a number of days adding or removing tracks to or from the library no longer works. Doing this will report success, but restarting psst will show the track is not added to or removed from "saved tracks".

What still halfway works is that when you leave psst-gui running, adding a track to the library will add it _invisibly_ to the list. The list will not _show_ the new track being there, but clicking on the top track in the list will play the newly added track. This bug has been there since forever. Previously that was always easily fixed by just restarting psst-gui - but now that no longer works because the list isn't actually saved.

**To Reproduce**
Add a track to, or remove a track from the Library (i.e. "Saved Tracks").

**Expected behavior**
The track is added to, or removed from "Saved Tracks".

**Screenshots**

**Environment**

- OS:
- Version:

**Additional context**

SO9010:
Fixes #669.

@jacksongoode could you quickly look though this.

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -113,7 +113,8 @@ impl WebApi {
                 .agent
                 .delete(request.build())
                 .header("Authorization", &format!("Bearer {}", token))
-                .call()
+                .force_send_body()
+                .send_json(request.get_body())
                 .map_err(|err| Error::WebApiError(err.to_string())),
         }
     }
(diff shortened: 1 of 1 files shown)


jacksongoode:
How did you discover that this works?

https://docs.rs/ureq/latest/ureq/struct.RequestBuilder.html#method.force_send_body

This seems like a workaround method? Are there other instances of the Spotify API requiring this on other projects?

jacksongoode (approved):
Seems good! Weird but confirmed with Spotify's API.

PR #680, opened 2025-07-17
Update build.yml

Cleboost:
Diff with mine (workin fork)


jacksongoode:
This won't produce the proper application this way. If you download the artifact of a build like this you will see it has no Psst.app

PR #681, opened 2025-07-23
feat: adjust volume mapping for linear perceived loudness

Guanran928:
This improves how the volume control works by making the loudness feel more natural to our ears.
Instead of changing volume linearly (which sounds uneven), it applies a logarithmic curve so that when you move the volume slider, the perceived loudness changes smoothly and evenly.


Guanran928:
Wait... It is already implemented in https://github.com/jpochyla/psst/pull/339

It's weird that the volume still sounds uneven to me.

PR #682, opened 2025-07-25
Detect System Color Theme

NandeMD:
This PR adds two things:
* On the first startup, (if there is no `$HOME/.config/Psst/config.json` file), app tries to detect the system theme, falls back to `Theme::Light` on error or unspecified theme.
* A new button on the `Preferences -> General Tab -> Theme`. When pressed, it tries to detect the system theme, falls back to `Theme::Light` on error or unspecified theme:
    
[Screencast_20250725_143112.webm](https://github.com/user-attachments/assets/830c1285-b21d-4289-9787-53842f18c856)


jacksongoode:
Super cool thank you! Did you see #162 and #160 and #150 ?

PR #693, opened 2025-09-21
Implement Login5 Authorization

Pogodaanton:
Retrieving an `access_token` over Keymaster broke a while ago. The librespot folks alleviated the auth question by [resorting to the Login5](https://github.com/librespot-org/librespot/issues/1532#issuecomment-3188123661) interface instead. This pull-request ports the necessary code for Auth5 to work in `psst-core`.

### Protobuf
The new Auth5 interface requires a lot of new protobuf messages. While we could plug them into `psst-protocol`, I realized that the current `.proto` files in psst haven't been touched for 5 years. I could not even get the `pb-rs` build script to work without having to edit the existing protobuf files. I'm sure I could eventually figure out the build process, however, Librespot already provides their protobuf code in a neat package (`librespot-protocol`). I propose slowly moving our code to use it instead.

### Dependencies
As I am quite a beginner regarding Rust, I have ported the code in a rudimentary way. I pulled in quite a few more dependencies from librespot to make the code work. We need to look into whether we can replace them with existing dependencies.

## PR State

The PR is in a reviewable state. All the features I wished to implement are included now. This includes:

- auth5 interface for retrieving access token via stored credentials
- client token requesting interface for auth5 and potential future features
- caching for auth5 and client tokens; cached tokens are refreshed after their predetermined expiry time

### TODOs
- [x] Real world testing on platforms other than Windows (need help)
- [x] Token expiry/refresh testing. I currently did not see the tokens expire even in long listening sessions.
- [ ] Assess new dependencies: Can we replace them with existing ones?

Fixes #687

PR #694, opened 2025-09-21
Fix depcreciated reliance on Keymaster tokens

jacksongoode:
- Make OAuth/PKCE primary
- Depreciate Keymaster
- Refresh tokens on demand

There are properly a handful of other things here. I didn't want to get rid of all the Keymaster stuff, but it might be that we should eventually do that too. I'd like to follow librespot on this and as I said before this is more of a bandaid. I think these problems ought to be resolved by integrating librespot entirely but that would be a big effort.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aes"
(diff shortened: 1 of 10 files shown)


RensOliemans:
We're a day further, presumably my token has expired, and when restarting psst I get 401 errors. Logging out and back in works (well psst complained that another service was listening on 8888, when I restarted psst it worked fine, but I'm not sure if this problem was on my side or not, I'll check that next time)

RensOliemans:
I downloaded the new binary from https://github.com/jpochyla/psst/actions/runs/17954044878 (built after 2f7c54397ae03a544d20cbc73c6e34fe976aabbd, I believe, `psst-x86_64-unknown-linux-gnu`, `sha256:e36e29a0a0aa84bc86df42dc1bd544f449c398e300639aeb046b611bcb43521a`) and the refresh token works perfectly, I did not have to re-authenticate.

psst-gui/src/token_utils.rs
```diff
@@ -0,0 +1,197 @@
...
+    ///   persisted refresh token unless `clear_refresh_if_none` is true (then
+    ///   it will be cleared).
+    pub fn persist_tokens(
+        config: &mut Config,
+        access_token: Option<&str>,
+        refresh_token: Option<&str>,
+        clear_refresh_if_none: bool,
+        save: bool,
+    ) {
+        log::info!("TokenUtils: persist save={save} clear_refresh_if_none={clear_refresh_if_none}");
```
RensOliemans, line 93: All of these TokenUtil logs seem better suited to debug, right? I guess it is useful information to see when the token is refreshed. Aside: how do I change the logging level when running the binary?

jacksongoode:
Yeah, I'm finding this as well. I don't know if this technique will actually work to keep refreshing for tokens... Grr. This approach could potentially be a dead end and login5 might be the solution?

arch-btw:
@jacksongoode feel free to ignore this if you don't think that's it but what about rotating the 6 access points that it fetches? They might be rate limited by access point. So for example if this one fails:

`trying AP 1 of 6: ap-gew4.spotify.com:4070`
`OAuth error: Failed to refresh token`

To try a different one?

jacksongoode:
Closed in favor of #693

PR #701, opened 2025-11-16
save_album: Move album id from json body to query parameter

### LINKED ISSUE #686: Cannot add or remove albums from library
**Describe the bug**
HTTP status 400

**To Reproduce**
Try to add or remove an album from your library

**Expected behavior**
It adds or removes the album from the library

**Screenshots**

**Environment**

- OS: macOS
- Version: 15.5

Image

Pogodaanton:
Turns out, Spotify's API wants the album ID that is being added to the user's library to be a query parameter instead of being part of the request body. Either they changed their mind or it always was the case, but never enforced it previously.

Fixes #686


jacksongoode:
Looks great! Thanks!

PR #702, opened 2025-11-23
Swap platform_dirs for directories for config paths

jacksongoode:
Old crate we can get rid of nicely.

### DIFF
--- a/psst-gui/src/data/migration.rs
+++ b/psst-gui/src/data/migration.rs
@@ -1,0 +1,109 @@
+use std::{fs, path::Path};
+
+use directories::ProjectDirs;
+use platform_dirs::AppDirs;
+
+use crate::data::config::Config;
+
+const APP_NAME: &str = "Psst";
+
+pub fn perform_migration() {
+    let old_app_dirs = AppDirs::new(Some(APP_NAME), false);
+    let new_project_dirs = ProjectDirs::from("", "", APP_NAME);
+
+    if let (Some(old_dirs), Some(new_dirs)) = (old_app_dirs, new_project_dirs) {
+        migrate_path(
+            &old_dirs.config_dir,
+            new_dirs.config_dir(),
+            &["config.json"],
+        );
+        migrate_path(
+            &old_dirs.cache_dir,
+            new_dirs.cache_dir(),
+            &[
+                "tracks",
+                "track",
+                "episodes",
+                "episode",
+                "audio",
+                "keys",
+                "key",
+                "lyrics",
+                "images",
+                "artist-info",
+                "related-artists",
+                "album",
+                "show",
+                "user-info",
+            ],
+        );
+    }
+
+    if let Some(active_cache_dir) = Config::cache_dir() {
+        rename_cache_subdirectories(&active_cache_dir);
+    }
+}
+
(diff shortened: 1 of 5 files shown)


Pogodaanton (changes requested):
Directory paths don't map 1:1 from `platform_dirs` to `directories` on Windows. E.g. former puts configs in the [application's roaming root directory](https://docs.rs/platform-dirs/latest/src/platform_dirs/lib.rs.html#185), while latter puts configs in a [sub-directory](https://docs.rs/directories/6.0.0/directories/struct.ProjectDirs.html#method.config_dir). Similar story to cache directories.

Given this, users get logged out on Windows, past settings are forgotten and past cached data require a manual clean by the user.

Pogodaanton:
I haven't found a standard for cache/config directories for Windows, except for the general advice to use AppData/Local or AppData/Roaming for user-specific miscellaneous application data (Local is fine for our use-case).
The `directories` library uses a [deliberately opinionated](https://github.com/dirs-dev/directories-rs/issues/9) folder structure for the various file groups (cache, config, etc.). They seem to be inspired by QT's conventions, though they even deviated from that.

Overall, I would say that if we opt for using `directories`, we should use their default folder structures, otherwise we can pick other less strict path resolving packages. A small benefit to sub-folders is that safe-to-delete folders like cached files will be easier to find.

A simple migration script should do the job, it can be deleted in a year or so I'd say.

psst-gui/src/data/migration.rs
```diff
@@ -0,0 +1,109 @@
...
+                "keys",
+                "key",
+                "lyrics",
+                "images",
+                "artist-info",
+                "related-artists",
+                "album",
+                "show",
+                "user-info",
+            ],
```
Pogodaanton:
```suggestion
                "user-info",
                "country_code",
            ],
```

Country code is cached by `psst-core`... Honestly it feels weird that it's a tiny 2B file cached like this, but that's a story for another day.

psst-gui/src/data/migration.rs
Pogodaanton: Maybe the file could be renamed, just in case. What if there is going to be a migration of another kind in the future, how are we going to differentiate it to this migration? IMO we can afford to call this file something verbose like `migration_paths_to_use_directories.rs` since we aim to remove this some time in the future eventually.

Pogodaanton (changes requested):
I tested the migration process on Windows. It worked, except for the missing `country_code` migration.

psst-gui/src/data/migration.rs
jacksongoode: Yeah. Hmm, I renamed legacy which might be enough? Everything in that module should eventually go away.

jacksongoode:
>  I tested the migration process on Windows. It worked, except for the missing country_code migration.


Thanks for the test! I need one more test on Linux to validate that it's all good.

Cleboost:
I'm on linux, what do I need to test?

Pogodaanton:
> I'm on linux, what do I need to test?

In theory, nothing should have changed on Linux/macOS. That's what needs to be verified.

A quick smoke test is to verify whether you're still logged in when using this branch and whether your settings are the same as in previous versions. In settings under the section "cache", the size value should approximately (!) match the size in previous versions.

PR #708, opened 2025-11-25
Add public user section feature.

SO9010:
This is the PR for the feature #616.

image

This is doing a similar thing to how the homepage works and should have essentially the same functionality as the Spotify one. Something that needs discussing is how can we should all the playlists or followers?


SO9010:
We should make a placeholder icon for playlists with out cover images

SO9010:
@jacksongoode I would like some input on the functionality please, currently, it just loads the first 10 for the artist and playlists, but we can get all of them, should we just load all of them in a scroll thing, or should we make it so we can say clikco nthe title to bring it to all of them?

Edit: I can change how many is gotten i think its cleaner if we leave it just showing 20 of them, but we should have a way to view all

SO9010:
Ok so I have set the unloadable playlist to the default playlist icon, I will do the same for the users. I will add (un)follow next, then I will request a review!!

SO9010:
Alright, I'm quite happy with where this is at now, @jacksongoode could you please look over the code as there is a lot of it XD

psst-gui/src/data/mod.rs
```diff
@@ -8,6 +8,7 @@ mod nav;
 mod playback;
 mod playlist;
 mod promise;
+pub mod public_user;
```
jacksongoode, line 11: Why does this need to be public?

psst-gui/src/data/mod.rs
```diff
@@ -170,6 +172,9 @@ impl AppState {
             finder: Finder::new(),
             lyrics: Promise::Empty,
             credits: None,
+            public_user_detail: PublicUserDetail {
+                info: Promise::Empty,
+            },
```
jacksongoode, line 177: Was there a reason to create a struct to hold only the info? Otherwise this could be just a Promise?

psst-gui/src/data/nav.rs
```diff
@@ -65,6 +67,7 @@ impl Nav {
             Nav::SavedAlbums => "Saved Albums".to_string(),
             Nav::Shows => "Podcasts".to_string(),
             Nav::SearchResults(query) => query.to_string(),
+            Nav::PublicUserDetail(usr) => usr.get_display_name().to_string(),
```
jacksongoode: We can probably use `user` :) just one more char.

jacksongoode: Also any reason we are using a getter here? Compared to just having it be a field like .name (like the others?)

psst-gui/src/data/user.rs
```diff
@@ -10,8 +10,48 @@ pub struct UserProfile {
...
+impl PublicUser {
+    /// Get the display name, falling back to `name` field if `display_name` is
+    /// empty
+    pub fn get_display_name(&self) -> Arc<str> {
+        if self.display_name.is_empty() {
+            self.name.clone().unwrap_or_default()
+        } else {
+            self.display_name.clone()
+        }
+    }
```
jacksongoode: So this can be optional? That's kind of unexpected, I would imagine this would be required?

jacksongoode: I guess as with id you are cloning the name if it doesn't exist? But this should be happening in the creation of a new user I think, not in the getter method dynamically?

psst-gui/src/ui/public_user.rs
```diff
@@ -0,0 +1,368 @@
...
+            |_, data, (_username, r)| {
+                if let Err(err) = r {
+                    data.error_alert(err);
+                } else {
+                    data.info_alert("User unfollowed.");
+                    // Update the is_following state in the cached data
+                    if let Some(cached) = data.public_user_detail.info.resolved_mut() {
+                        let info = Arc::make_mut(&mut cached.data);
+                        info.is_following = Some(false);
+                        info.followers_count -= 1;
```
jacksongoode, line 79: I guess it makes sense that we update these values, but do we ever actually refresh these values elsewhere? Not that it matters, just curious. I'm trying to think about how Psst might cache things, but some fields shouldn't really need to be cached if they are dynamic, like follower counts, listener counter perhaps.

psst-gui/src/ui/public_user.rs
```diff
@@ -0,0 +1,368 @@
...
+                    }
+                }
+            },
+        )
+}
+
+fn loaded_detail_widget() -> impl Widget<WithCtx<Cached<Arc<PublicUserInformation>>>> {
+    let user_profile_top = user_info_widget().padding(theme::grid(1.0));
+
+    // This puts it as read only, it is needed to trainsform the context into what
```
jacksongoode:
```suggestion
    // This puts it as read only, it is needed to transform the context into what
```

psst-gui/src/ui/public_user.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    // Placeholder widget showing a user icon
+    let placeholder = icons::USER
+        .scale((size * 0.9, size * 0.9))
+        .with_color(Color::rgb8(0x53, 0x53, 0x53))
+        .padding((size * 0.05, size * 0.05))
+        .center()
+        .fix_size(size, size)
+        .background(Color::rgb8(0x28, 0x28, 0x28))
+        .clip(Circle::new((radius, radius), radius));
```
jacksongoode: Might be good to put this in our icons folder and then import? Maybe there is a nice SF Pro icon we can use?

psst-gui/src/ui/public_user.rs
```diff
@@ -0,0 +1,368 @@
...
+                ctx.data.data.image_url.clone().map(|url| url.into())
+            },
+        )
+        .fix_size(size, size)
+        .clip(Circle::new((radius, radius), radius)),
+        placeholder,
+    )
+}
+
+fn user_info_widget() -> impl Widget<WithCtx<Cached<Arc<PublicUserInformation>>>> {
```
jacksongoode, line 154: It'd be great to have a photo on the PR description too!

psst-gui/src/ui/public_user.rs
```diff
@@ -0,0 +1,368 @@
...
+                    .padding((theme::grid(1.5), theme::grid(0.5))),
+            )
+            .with_child(
+                Scroll::new(List::new(|| artist::artist_widget(true)).horizontal())
+                    .horizontal()
+                    .align_left()
+                    .lens(Ctx::data().then(MixedView::artists)),
+            ),
+    )
+}
```
jacksongoode, line 259: These can be a widget builder function to make these widgets with diff content and titles since they use all the same padding, margins etc.

psst-gui/src/webapi/client.rs
```diff
@@ -688,6 +692,281 @@ impl WebApi {
...
+            pub followers_count: i64,
+            pub image_url: String,
+            #[serde(default)]
+            pub is_following: bool,
+            pub name: String,
+            pub uri: String,
+        }
+
+        #[derive(Clone, Data, Deserialize)]
+        pub struct PublicUserPlaylist {
```
jacksongoode, line 717: Just curious, how is this different than a playlist?

psst-gui/src/webapi/client.rs
```diff
@@ -688,6 +692,281 @@ impl WebApi {
...
+            #[serde(default)]
+            pub following_count: i64,
+            #[serde(default)]
+            pub is_following: Option<bool>,
+            #[serde(default)]
+            pub is_current_user: Option<bool>,
+            #[serde(default)]
+            pub recently_played_artists: Vector<PublicUserArtist>,
+            #[serde(default)]
+            pub public_playlists: Vector<PublicUserPlaylist>,
```
jacksongoode, line 743: Like perhaps this should just be a vec of playlists?

psst-gui/src/webapi/client.rs
```diff
@@ -688,6 +692,281 @@ impl WebApi {
...
+            is_current_user: result.is_current_user,
+            recently_played_artists: artist,
+            public_playlists: playlist,
+            allow_follows: result.allow_follows,
+            followers,
+            following,
+        })
+    }
+
+    pub fn get_public_user_followers(&self, id: Arc<str>) -> Result<Vector<PublicUser>, Error> {
```
jacksongoode, line 860: I think some of these UI widgets can be abstracted here too.

psst-gui/src/webapi/client.rs
```diff
@@ -688,6 +692,281 @@ impl WebApi {
...
+    pub fn follow_user(&self, username: &str) -> Result<(), Error> {
+        let json = json!({
+            "variables": {
+                "usernames": [username]
+            },
+            "operationName": "followUsers",
+            "extensions": {
+                "persistedQuery": {
+                    "version": 1,
+                    "sha256Hash": "c00e0cb6c7766e7230fc256cf4fe07aec63b53d1160a323940fce7b664e95596"
```
jacksongoode: Can we store this hash somewhere as a const and import it? Hmm I'm just thinking future wise it would be good to have them stored somewhere.

psst-gui/src/ui/public_user.rs
jacksongoode: Oh you did, but maybe we can abstract this icon builder chain.

jacksongoode (changes requested):
Nice it looks great! Here's my first pass of comments.

psst-gui/src/data/nav.rs
SO9010: The issue is the inconsistencies with the serialisation of public user, sometimes it is display_name and sometimes its name...

SO9010:
Thanks for the review, I will go through this in a bit!

psst-gui/src/webapi/client.rs
SO9010: I like this idea a lot perhaps we could have it at the top?

psst-gui/src/data/mod.rs
SO9010: Because we use it nav.rs and client.rs without it being public it is unaccessible :)

psst-gui/src/webapi/client.rs
SO9010: Alright I have done something which I think works

psst-gui/src/webapi/client.rs
SO9010:
``` 
PublicUserPlaylist {
        pub image_url: String,
        pub name: String,
        pub owner_name: String,
        pub owner_uri: String,
        pub uri: String,
        #[serde(default)]
        pub followers_count: Option<i64>,
        #[serde(default)]
        pub is_following: Option<bool>,
}
```


vs
```
Playlist {
        pub id: Arc<str>,
        pub name: Arc<str>,
        #[serde(skip_serializing_if = "Option::is_none")]
        pub images: Option<Vector<Image>>,
        #[serde(deserialize_with = "deserialize_description")]
        pub description: Arc<str>,
        #[serde(rename = "tracks")]
        #[serde(deserialize_with = "deserialize_track_count")]
        pub track_count: Option<usize>,
        pub owner: PublicUser,
        pub collaborative: bool,
        #[serde(rename = "public")]
        pub public: Option<bool>,
}
```

psst-gui/src/ui/public_user.rs
SO9010: No we don't refresh them elsewhere :P, not actually sure why I did this like this anymore... Do you think I should change it?

psst-gui/src/data/mod.rs
SO9010: This is in case we wanted to add the ability to view all for different elements, so for example you can have all of the user playlists or artists or friends there, and that imo would be better split into different parts?

SO9010:
Annoyingly psst isnt loading at all anymore, the program runs for me on all builds, but then the whole ui crashes... I hope your not getting the same issue...

Edit: Just needed a reboot XD

psst-gui/src/webapi/client.rs
SO9010: There are a lot of differences, mainly the fact that the public user part is expanded, I can reduce the side of the PublicUserPlaylist though!

psst-gui/src/ui/public_user.rs
SO9010: I think i have done that now if i understand you correctly!

SO9010:
@jacksongoode I was looking into lazy scrolling with druid, and unfortunately, it be possible, at least in a clean way, there is no way I belive to append onto a list view...

psst-gui/src/webapi/client.rs
jacksongoode: I guess what I meant here is that `get_public_user_followers`, `get_public_user_following` are identical except for the request url and could be abstracted.

psst-gui/src/webapi/client.rs
```diff
@@ -1121,6 +1397,15 @@ impl WebApi {
...
 /// View endpoints.
 impl WebApi {
+    const HOME_SECTION_MADE_FOR_YOU: &str = "spotify:section:0JQ5DAUnp4wcj0bCb3wh3S";
+    const HOME_SECTION_TOP_MIXES: &str = "spotify:section:0JQ5DAnM3wGh0gz1MXnu89";
+    const HOME_SECTION_RECOMMENDED_STATIONS: &str = "spotify:section:0JQ5DAnM3wGh0gz1MXnu3R";
+    const HOME_SECTION_UNIQUELY_YOURS: &str = "spotify:section:0JQ5DAUnp4wcj0bCb3wh3S";
+    const HOME_SECTION_BEST_OF_ARTISTS: &str = "spotify:section:0JQ5DAnM3wGh0gz1MXnu3n";
+    const HOME_SECTION_JUMP_BACK_IN: &str = "spotify:section:0JQ5DAIiKWzVFULQfUm85X";
+    const HOME_SECTION_YOUR_SHOWS: &str = "spotify:section:0JQ5DAnM3wGh0gz1MXnu3N";
+    const HOME_SECTION_SHOWS_YOU_MIGHT_LIKE: &str = "spotify:section:0JQ5DAnM3wGh0gz1MXnu3P";
```
jacksongoode: Nice, thanks!

Issue #624, opened 2025-06-09
Add URL encoding to search


SO9010:
Currently, we are just using a placeholder to replace the spaces in a query but it needs to changed to use URL encoding.

arch-btw:
Hmm, do you mean that this line needs to be improved upon, or are you referring to something else?

https://github.com/jpochyla/psst/blob/0e9c8d0283ced2c453aafa836be16de9fd51e48c/psst-gui/src/webapi/client.rs#L1351

SO9010:
Yeah that line!

SO9010:
I believe that if we don't do url encoding then we will get errors from illegal characters in urls. Which is why we needed to replace the spaces with %20. But I believe that we also need to do things like ':'. 

But it was just a quick submitted issue that I didn't 100 percent check. 

So if you would like to you could check if a query with something like : or any other not allowed chars in a url please feel free to do so. If not I'll be able to test soon :)

SO9010:
Wow thank you very much!

PR #710, opened 2025-11-28
Search by type

jacksongoode:
Search by types! It will save the last category you searched by for your next search! I've limited it to 50 but I think in a future PR we could expose it to the config page so user's could decide how many results they want.

<video width="400" src="https://github.com/user-attachments/assets/26c3f276-9e55-4e50-83b9-643417b1a1ba" ></video>

### DIFF
--- a/psst-gui/src/controller/nav.rs
+++ b/psst-gui/src/controller/nav.rs
@@ -31,8 +31,14 @@ impl NavController {
             Nav::SearchResults(query) => {
                 if let Some(link) = SpotifyUrl::parse(query) {
                     ctx.submit_command(search::OPEN_LINK.with(link));
-                } else if !data.search.results.contains(query) {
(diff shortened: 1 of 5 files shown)


jacksongoode:
I suppose its a bit redundant to show the headers of the section if you have already selected by it...

SO9010:
Yeah I agree, if its selected we shouldnt show the it as a header

SO9010:
IMO the code looks good, nothing which jumps out at me... Ill give it a test in a second!

SO9010:
We should possibly add a fix for #624 in this? We whoucl also add something like duckduckgo bangs to make it quicker to search for specific thigns like !artist super cool and that would just search the artists and go to the tab?

jacksongoode:
> We should possibly add a fix for https://github.com/jpochyla/psst/issues/624 in this?

Maybe let's do it in a separate PR!

PR #711, opened 2025-11-30
Fix truncation on playlist descriptions & album/search titles

jacksongoode:
Before long titles or descriptions would not respect the gutters/padding:

imageimage

---

Now they do, it also means the sorting is anchored right (which I think is fine):

image


jacksongoode:
Going to go ahead with this one because it's harmless.

PR #714, opened 2025-12-03
Replace custom protobuf handlers with librespot-protocol

Pogodaanton:
In #693 we have introduced `librespot-protocol` and its direct dependency `protobuf` to `psst-core`. The code now uses two different protobuf libraries. This PR aims to reduce it back to one.

I have been unsuccessful at porting new protobuf specs to `quick-protobuf`. I can't get the bash scripts in `psst-protocol` running, nor could I reproduce the rust code generation for the already existing protobufs. I tried tackling the problem from the other side instead: Use more of the librespot protobuf code.


psst-core/src/connection/mod.rs
```diff
@@ -232,17 +231,18 @@ impl Transport {
         // hashed together with the shared secret to make a key pair).
         log::trace!("waiting for AP response");
         let apresp_packet = read_packet(&mut stream)?;
-        let apresp: APResponseMessage = deserialize_protobuf(&apresp_packet[4..])?;
+        let apresp = APResponseMessage::parse_from_bytes(&apresp_packet[4..])?;
         log::trace!("received AP response");
 
         // Compute the challenge response and the sending/receiving keys.
-        let remote_key = &apresp
```
SO9010, line 239: Any reason for changing this?

SO9010 (commented):
Looks great to me, I need @jacksongoode to look over this as I have not been active for a while... Thanks for the help!

psst-core/src/connection/mod.rs
Pogodaanton: I'm using `as_ref` in line 243, so the resulting value is still a reference. I did this because values are boxed in the `protobuf` library, so it made sense to unwrap the `Option` later in the chain.

jacksongoode (commented):
Generally this looks fine, does this mean we can remove a lot of the native code from protocol?

Pogodaanton:
> Generally this looks fine, does this mean we can remove a lot of the native code from protocol?

It means that we can remove the entire folder `psst-protocol`. As I'm writing this, I've noticed that I haven't done so yet in this PR, however, there is no dependence on the `psst-protocol` package anywhere in the code anymore.

jacksongoode:
I would say go ahead, it makes sense to if all functionality appears to work.

jacksongoode:
I've gone ahead and done this. All appears to work. Given that this is a huge chunk of removed code, I'd like to see how it might change the package size.

Pogodaanton:
> I've gone ahead and done this. All appears to work. Given that this is a huge chunk of removed code, I'd like to see how it might change the package size.

Boy, do I have meager news for you 😅 

Jokes aside, the size won't change that much given that all the protocols removed are still used, only from a different package now. As far as I can see in the CI builds, the size change is in margin of error, effectively 0.

jacksongoode:
Let's do it, I think the convergence with librespot is likely the best path towards sustainability anyways.

PR #717, opened 2025-12-16
fix: use the urlencoding crate instead of manual replacement of special characters

### LINKED ISSUE #624: Add URL encoding to search

vitosotdihaet:
Closes #624, as suggested in https://github.com/jpochyla/psst/issues/624#issuecomment-2966336601

It seems there's no other place that needs this encoding

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4004,6 +4004,7 @@ dependencies = [
  "time-humanize",
  "ureq 3.0.11",
  "url",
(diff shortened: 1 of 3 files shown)


jacksongoode (approved):
Looks fine, urlencoding is ~6KiB with no dependencies.

Thank you!

PR #720, opened 2026-02-21
feat: use separate user-provided Client ID for Web API calls

grenaad:
## Disclosure
All code is generated with Opencode with Opus model. I tested this with my own client id, and auth is working. I have not used Rust in a few years, so I can't verify that the code quality is up to standard. I am happy to make any changes requested.
## Summary
Adds a separate OAuth2 PKCE flow for Web API calls (search, library, playlists) using a user-registered Spotify Developer Client ID. This avoids 429 rate-limiting from sharing the official client ID across all psst users.
Depends on #722  (spclient CDN switch).
**Setup:** Register a Spotify app at https://developer.spotify.com/dashboard with redirect URI `http://127.0.0.1:8888/login`, then enter the Client ID on the sign-in screen.
## Changes
- Users enter their Spotify Developer Client ID on the sign-in screen, persisted in `config.json`
- During login, a second OAuth2 PKCE flow runs after session auth (two browser redirects)
- `WebApiToken` (access + refresh + expiry) persisted in `config.json`
- Token auto-refreshes on startup and on expiry; cleared on logout
- `WebApi` no longer depends on `Login5`/`SessionService` — manages its own token via `set_webapi_credentials()`
- Web API OAuth failure is non-fatal: playback still works
- OAuth helpers deduplicated into parameterized shared functions


psst-core/Cargo.toml
```diff
@@ -19,7 +19,9 @@ log = { version = "0.4.22" }
 num-bigint = { version = "0.4.6", features = ["rand"] }
 num-traits = { version = "0.2.19" }
 oauth2 = { version = "4.4.2" }
+open = { version = "5.3.2" }
 parking_lot = { version = "0.12.3" }
+platform-dirs = { version = "0.3.0" }
```
Pogodaanton: I'd be wary of adding dependencies like `open` and `platform-dirs` to psst-core, as we already have those in psst-gui. Code that uses these dependencies should probably be in psst-gui.

psst-gui/src/ui/preferences.rs
```diff
@@ -585,27 +589,50 @@ impl Authenticate {
...
+                }
+
+                Ok(credentials)
             },
             Self::SPOTIFY_RESPONSE,
             self.spotify_thread.take(),
         );
 
-        // Open browser with authorization URL
+        // Open browser with authorization URL (for the first/session OAuth)
```
Pogodaanton, line 710: RE `open` dependency in psst-core: This is probably the place where the psst-core code using `open` should actually be at.

psst-core/src/session/access_token.rs
```diff
@@ -8,11 +8,38 @@ use crate::error::Error;
...
 // Client ID of the official Web Spotify front-end.
+// Used for Shannon session, Login5, client tokens, and playback infrastructure.
 pub const CLIENT_ID: &str = "65b708073fc0480ea92a077233ca87bd";
 
-// All scopes we could possibly require.
+// Psst's own registered Client ID for Web API calls.
+// This avoids rate-limiting (429 errors) that Spotify applies to the official client ID
+// when used for JSON Web API requests.
+// TODO: Register your own client ID at https://developer.spotify.com/dashboard and replace this.
+pub const PSST_CLIENT_ID: &str = "TODO_REGISTER_YOUR_OWN_CLIENT_ID";
```
Pogodaanton: This field needs to be exposed into the GUI, otherwise everyone will have to compile psst by themselves.

psst-gui/src/main.rs
```diff
@@ -62,6 +62,29 @@ fn main() {
     let launcher;
     if state.config.has_credentials() {
         // Credentials are configured, open the main window.
+
+        // Check if we have a valid Web API token. If not, try to refresh or
+        // trigger a browser-based OAuth flow before the main window opens.
+        match oauth::get_or_refresh_webapi_token() {
```
Pogodaanton: In my limited tests (re-opened this build a day later and watched the logs), token refreshing seems to work!

psst-core/src/cdn.rs
```diff
@@ -29,41 +34,55 @@ impl Cdn {
             session,
             agent: agent.into(),
             login5: Login5::new(None, proxy_url),
+            client_token_provider: ClientTokenProvider::new(proxy_url),
```
Pogodaanton: Login5 also uses a client token provider. If we create yet another one, then we might be unnecessarily doing round-trip calls to the same API to get the same result. We should consider reusing the same ClientTokenProvider instance in multiple places.

psst-core/src/cdn.rs
```diff
@@ -29,41 +34,55 @@ impl Cdn {
...
+        let client_token = self.client_token_provider.get()?;
+        let mut response = self
             .agent
             .get(&locations_uri)
-            .query("version", "10000000")
-            .query("product", "9")
-            .query("platform", "39")
-            .query("alt", "json")
-            .header("Authorization", &format!("Bearer {}", access_token.access_token))
+            .header(
```
Pogodaanton, line 79: Any reason why the old query headers aren't sent anymore?

psst-core/src/oauth.rs
```diff
@@ -196,3 +203,256 @@ fn get_scopes() -> Vec<Scope> {
...
+    BasicClient::new(
+        ClientId::new(PSST_CLIENT_ID.to_string()),
+        None,
+        AuthUrl::new("https://accounts.spotify.com/authorize".to_string()).unwrap(),
+        Some(TokenUrl::new("https://accounts.spotify.com/api/token".to_string()).unwrap()),
+    )
+    .set_redirect_uri(
+        RedirectUrl::new(webapi_redirect_uri(redirect_port)).expect("Invalid redirect URL"),
+    )
+}
```
Pogodaanton: This is almost the same code as in line 158. I understand the need to keep the systems separate, but it feels a bit repetitive to define the same URLs and clients (barring different clientids) twice.

psst-core/src/oauth.rs
```diff
@@ -196,3 +203,256 @@ fn get_scopes() -> Vec<Scope> {
...
+    let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
+
+    let (auth_url, _) = client
+        .authorize_url(CsrfToken::new_random)
+        .add_scopes(get_webapi_scopes())
+        .set_pkce_challenge(pkce_challenge)
+        .url();
+
+    (auth_url.to_string(), pkce_verifier)
+}
```
Pogodaanton: Code here repeats the above (line 171) as well

psst-core/src/oauth.rs
```diff
@@ -196,3 +203,256 @@ fn get_scopes() -> Vec<Scope> {
...
+
+    // Exchange code for token
+    let token = exchange_webapi_code_for_token(redirect_port, code, pkce_verifier)?;
+
+    // Save to disk
+    save_webapi_token(&token)?;
+
+    log::info!("Web API OAuth flow completed successfully");
+    Ok(token)
+}
```
Pogodaanton: RE comment in `cargo.toml`: Psst already does credential persistence in psst-gui (that's why desktop client id credentials aren't persisted in this file). There's also a config file psst-gui manages in which the shannon credentials are stored. This code skips both of that. Moving this code to psst-gui would alleviate adding two new dependencies to psst-core.

Pogodaanton (commented):
Tried it with a custom client id and it worked on my end, too. 
Until the maintainers review this, I'd like to share some thoughts regarding this PR.

In general, I support the authentication approach implemented here. Adding an endpoint resolver makes sense. There is some code that I'd move to other places - I've pinned a comment on those. After that, the next step should be exposing the client id as an input field on sign-in (instead of hardcoding it on build) and explaining to the user that there are **two** authentication steps.

As the developer API also changed [quite a lot](https://developer.spotify.com/documentation/web-api/references/changes/february-2026), we also need to update many API requests. I don't know whether this should be a separate PR or not (I lean towards saying yes).

Either way, I appreciate your help and the quick PR!

psst-core/Cargo.toml
grenaad: Agreed, removed both `open` and `platform-dirs` from `psst-core`. All code using them now lives in `psst-gui`.

psst-gui/src/ui/preferences.rs
grenaad: Done, browser-opening for the Web API OAuth flow now happens directly in `preferences.rs` within the auth thread.

psst-core/src/session/access_token.rs
grenaad: Removed the hardcoded constant entirely. The client ID is now a user-facing text input on the sign-in screen, persisted in `config.json`.

psst-gui/src/main.rs
grenaad: Good to hear. Token refresh is now managed from `psst-gui` via `Config::get_or_refresh_webapi_token()`, which saves back to config on successful refresh.

psst-core/src/cdn.rs
grenaad: Fixed, `Cdn::new()` now creates a single `Arc<ClientTokenProvider>` shared between Login5 and Cdn, eliminating the redundant round-trip.

psst-core/src/cdn.rs
grenaad: The `spclient` endpoint returns protobuf natively and doesn't need `?alt=json&version=...&product=...&platform=...` , those were specific to the old `api.spotify.com` JSON endpoint. Matches `librespot's` implementation. Added a comment explaining this.

psst-core/src/oauth.rs
grenaad: Consolidated into a single `create_oauth_client(client_id, redirect_port)` and `generate_auth_url(client_id, redirect_port, scopes)` used by both session and Web API flows.

psst-core/src/oauth.rs
grenaad: Same as above, both flows now call the shared `generate_auth_url()` with different parameters.

psst-core/src/oauth.rs
grenaad: Moved entirely to `psst-gui`. Token is now stored inside `config.json` alongside existing credentials via `Config::store_webapi_token()` / `get_or_refresh_webapi_token()`. Removed `open` and `platform-dirs` from `psst-core`.

Pogodaanton:
Please verify if the sign-in flow works in your latest commit. It did not on my end.

grenaad:
> Please verify if the sign-in flow works in your latest commit. It did not on my end.

I have verified it, I did had to sign out and sign in again. Did the auth fail for you?

jacksongoode:
Thank you for the PR, sorry I haven't been following Spotify's recent API changes but just to confirm, @grenaad @Pogodaanton is this the approach that is used by librespot?

Pogodaanton:
> Thank you for the PR, sorry I haven't been following Spotify's recent API changes but just to confirm, @grenaad @Pogodaanton is this the approach that is used by librespot?

Yes, this PR aligns our CDN endpoint system with librespot's, such that we [don't use](https://github.com/jpochyla/psst/pull/720/changes#diff-3fe25c524d5d8e5d44e50cdaa01ccd327a5927605f7c3e238bffaf1b50257e49R136) a hardcoded endpoint anymore. 

That was a relatively minor fix though, most of what this PR aims to do is to introduce a _second_ client-id that can be used for all the frontend API requests. The `api.spotify.com` API itself was [also changed](https://developer.spotify.com/documentation/web-api/references/changes/february-2026), so many of our current API requests in the frontend are borked and need updating. This PR has not addressed that issue yet (nor does it need to IMO as said previously).

psst-core/src/oauth.rs
```diff
@@ -12,14 +13,16 @@ use std::{
 };
 use url::Url;
 
+use crate::session::access_token::WEBAPI_SCOPES;
+
+// ── Callback listener (shared by all OAuth flows) ──────────────────────────
```
jacksongoode: No need to add these doc separators.

psst-core/src/oauth.rs
```diff
@@ -12,14 +13,16 @@ use std::{
...
+
 pub fn listen_for_callback_parameter(
     socket_address: SocketAddr,
     timeout: Duration,
     parameter_name: &'static str,
 ) -> Result<String, Error> {
-    log::info!(
-        "starting callback listener for '{parameter_name}' on {socket_address:?}",
-    );
+    log::info!("starting callback listener for '{parameter_name}' on {socket_address:?}",);
```
jacksongoode: No need for formatting changes here, the less diff the better so we can review the essentials.

psst-core/src/oauth.rs
```diff
@@ -148,38 +151,38 @@ pub fn send_success_response(stream: &mut TcpStream) {
...
 
-fn create_spotify_oauth_client(redirect_port: u16) -> BasicClient {
-    let redirect_address = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), redirect_port);
-    let redirect_uri = format!("http://{redirect_address}/login");
+// ── Shared OAuth helpers ───────────────────────────────────────────────────
+
+fn redirect_uri(redirect_port: u16) -> String {
+    let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), redirect_port);
+    format!("http://{addr}/login")
+}
```
jacksongoode: Helpers like this are nice but in this PR we are solving one specific issue here and this helper isn't needed for that. Later we can add in a follow PR.

Pogodaanton:
> > Please verify if the sign-in flow works in your latest commit. It did not on my end.
> 
> I have verified it, I did had to sign out and sign in again. Did the auth fail for you?

I tested this on Windows 11 via Firefox.

I cannot press the "Log in with Spotify" button in the setup workflow because the window cuts it off. I quickly changed the window size in the code, but after sign-in, the WebAPI OAuth handler errored out:

```
[2026-02-25T14:41:58Z INFO  psst_gui::ui::preferences] Session auth complete. Starting Web API OAuth flow...
[2026-02-25T14:41:59Z INFO  psst_core::oauth] starting callback listener for 'code' on 127.0.0.1:8888
[2026-02-25T14:41:59Z INFO  psst_core::oauth] listener bound successfully
[2026-02-25T14:41:59Z ERROR psst_core::oauth] Failed to extract parameter 'code' from request: GET /favicon.ico HTTP/1.1

[2026-02-25T14:41:59Z WARN  psst_gui::ui::preferences] Web API OAuth callback failed (non-fatal): OAuth error: Failed to extract parameter 'code' from request: GET /favicon.ico HTTP/1.1
```

It seems like the browser is attempting to fetch a favicon for the previous OAuth callback URL and our callback listener does not differentiate between different URL paths and assumes that any request to the socket address will contain an OAuth code.

psst-core/src/oauth.rs
```diff
@@ -190,9 +193,138 @@ pub fn exchange_code_for_token(
     token_response.access_token().secret().to_string()
 }
 
-fn get_scopes() -> Vec<Scope> {
+fn get_session_scopes() -> Vec<Scope> {
     crate::session::access_token::ACCESS_SCOPES
         .split(',')
         .map(|s| Scope::new(s.trim().to_string()))
         .collect()
 }
```
jacksongoode, line 195: For these I understand that they are now per session but lets keep them the same function names if we can so we can observe the changes inside them which would necessitate a naming change.

jacksongoode (changes requested):
Just some super style-wise comments which will help me review in a second round. I think we can make sure the changes here are very specific to the issue we want to address, later we can decide if the changes to add this client ID are good.

I am thinking that we should probably split this PR so the the switch to spclient would be a small one (even if the app is still broken).

psst-core/src/oauth.rs
grenaad: Reverted log::info! back to the original multi-line format.

psst-core/src/oauth.rs
grenaad: Removed all section separator comments

psst-core/src/oauth.rs
grenaad: Removed `redirect_uri()` helper and inlined the logic back into `create_oauth_client()`

jacksongoode:
@grenaad This is looking good, but would you be able to split this such that we have a PR just for the CDN change first? And then rebase this after we merge that so we can just check our the client ID separation on its own?

juanmalopman:
> I tested this on Windows 11 via Firefox.

> I cannot press the "Log in with Spotify" button in the setup workflow because the window cuts it off. I quickly changed the window size in the code, but after sign-in, the WebAPI OAuth handler errored out...

Same here. I first changed the window size using winspy.exe but now just changed the size in "Psst\psst-gui\src\ui\mod.rs" line 95 to just make the window "big":

```
pub fn account_setup_window() -> WindowDesc<AppState> {
    let win = WindowDesc::new(account_setup_widget())
        .title("Login")
        .window_size((theme::grid(80.0), theme::grid(95.0)))
        .resizable(false)
        .show_title(false)
        .transparent_titlebar(true);
    if cfg!(target_os = "macos") {
        win.menu(menu::main_menu)
    } else {
        win
    }
}
```

For the OAuth crash, I asked an AI (I've never programmed in Rust) and it turns out the bug happens because the local server was only accepting exactly one incoming connection. If the browser happens to send a background request for favicon.ico first, the server processes it, fails to find the Spotify code, throws a fatal error, and shuts down the listener completely before the real login redirect arrives.
To fix it, I changed the code in psst-core/src/oauth.rs (around line 43) to loop over incoming connections instead:

```
    // 3. Spawn the thread
    let handle = std::thread::spawn(move || {
        for stream in listener.incoming() {
            if let Ok(mut stream) = stream {
                // If the handler returns true, we got the code, so stop listening.
                if handle_callback_connection(&mut stream, &tx, parameter_name) {
                    break;
                }
            }
        }
    });
```

..and then updated handle_callback_connection (around line 73) to return a bool, ignore the favicon, and safely skip bad requests instead of crashing:

```
/// Handles an incoming TCP connection for a generic OAuth callback.
fn handle_callback_connection(
    stream: &mut TcpStream,
    tx: &mpsc::Sender<Result<String, Error>>,
    parameter_name: &'static str,
) -> bool {
    let mut reader = BufReader::new(&mut *stream);
    let mut request_line = String::new();

    if reader.read_line(&mut request_line).is_ok() {
        // Ignore background browser requests like favicon
        if request_line.contains("favicon.ico") {
            return false;
        }

        match extract_parameter_from_request(&request_line, parameter_name) {
            Some(value) => {
                log::info!("received callback parameter '{parameter_name}'.");
                send_success_response(stream);
                let _ = tx.send(Ok(value));
                return true; // Success! Tell the loop to break.
            }
            None => {
                let err_msg = format!(
                    "Failed to extract parameter '{parameter_name}' from request: {request_line}",
                );
                log::warn!("{err_msg}");
                // DO NOT send an error, just ignore the bad request and keep listening.
                return false;
            }
        }
    }
    false
}
```

After making these changes, `cargo build --release --bin psst-gui` compiled a working GUI and I was able to log in flawlessly!

grenaad:
> @grenaad This is looking good, but would you be able to split this such that we have a PR just for the CDN change first? And then rebase this after we merge that so we can just check our the client ID separation on its own?

#722 I created this for spclient CDN switch, after that has being merge, then this pr can be reviewed.

Cleboost:
Any news?

Cleboost:
is that ready?

PR #722, opened 2026-03-02
Switch CDN audio resolution from api.spotify.com to spclient

grenaad:
## Disclosure
All code is generated with Opencode with Opus model. I tested this with my own client id, and auth is working. I have not used Rust in a few years, so I can't verify that the code quality is up to standard. I am happy to make any changes requested.
## Summary
Spotify started returning 429 (Too Many Requests) for requests to `api.spotify.com/v1/storage-resolve`. This switches audio file URL resolution to Spotify's `spclient` infrastructure, matching librespot's approach.
Inspired by [ncspot commit 6769b7b](https://github.com/hrkfdn/ncspot/commit/6769b7b). Split from #720 per review feedback — Web API Client ID changes will follow in a subsequent PR.
## Changes
- Add `Transport::resolve_spclient()` to discover spclient access points via `apresolve.spotify.com/?type=spclient`
- `Cdn::resolve_audio_file_url()` now uses `https://{spclient}/storage-resolve/...` instead of `api.spotify.com/v1/storage-resolve`
- Parse protobuf `StorageResolveResponse` instead of JSON
- Send `client-token` header required by spclient endpoint
- Share `Arc<ClientTokenProvider>` between `Login5` and `Cdn` to avoid redundant round-trips
- Cache resolved spclient base URL for the `Cdn` instance lifetime

### DIFF
--- a/psst-core/src/cdn.rs
+++ b/psst-core/src/cdn.rs
@@ -40,7 +40,7 @@ impl Cdn {
         Ok(Arc::new(Self {
             session,
             agent: agent.into(),
-            login5: Login5::new(Some(client_token_provider.clone()), proxy_url),
+            login5: Login5::new(Some(Arc::clone(&client_token_provider)), proxy_url),
             client_token_provider,
             spclient_base: Mutex::new(None),
             proxy_url: proxy_url.map(String::from),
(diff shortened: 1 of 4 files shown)


jacksongoode:
Looking at this now 👀

jacksongoode (approved):
I'd rather get this in and get Psst back to working condition than nit pick. But just added two small edits. One was to abstract the resolve list creation that happens now in the ap and spclient methods, the other was an clone -> arc clone.

PR #730, opened 2026-05-12
fix: resolve #595 — Save the volume level

### LINKED ISSUE #595: Save the volume level
Every time I open the app, the volume is reset to 100%. It would be great if the application could remember the last volume level used and restore it automatically on the next launch.

chinhkrb113:
## Summary

The application's configuration struct (which is serialized/deserialized to disk) needs a field to store the volume level so it persists across launches. This file likely contains the `Config` struct with serde derives that gets saved to a config file

Fixes #595

## Changes

- `psst-gui/src/data/config.rs`

## Why

`psst-gui/src/data/config.rs`: The application's configuration struct (which is serialized/deserialized to disk) needs a field to store the volume level so it persists across launches. This file likely contains the `Config` struct with serde derives that gets saved to a config file.

### DIFF
--- a/psst-gui/src/data/config.rs
+++ b/psst-gui/src/data/config.rs
@@ -21,6 +21,10 @@ use serde::{Deserialize, Serialize};
 use super::{Nav, Promise, QueueBehavior, SliderScrollScale};
 use crate::ui::theme;
 
+fn default_volume() -> f64 {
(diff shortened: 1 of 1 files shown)


jacksongoode:
The volume already persists across sessions? What was this PR intended to do?

PR #733, opened 2026-06-27
Quality: OAuth callback listener thread and TCP port leak on timeout

kumburovicbranko682-boop:
## Problem

When `listen_for_callback_parameter` times out via `rx.recv_timeout`, the function returns early with an error at step 4, skipping `handle.join()` at step 5. The spawned thread remains blocked on `listener.incoming()` (or on a slow `read_line`) indefinitely. Because the `TcpListener` is owned by the thread and never dropped, the port stays bound until the process exits. Any subsequent OAuth callback attempt that tries to bind the same port will fail with "address already in use", making the retry path dead.

**Severity**: `medium`
**File**: `psst-core/src/oauth.rs`

## Solution

Store the listener handle and forcibly clean it up on timeout. The simplest fix is to make the thread joinable even in the error path:

## Changes

- `psst-core/src/oauth.rs` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/psst-core/src/oauth.rs
+++ b/psst-core/src/oauth.rs
@@ -64,6 +64,12 @@ pub fn listen_for_callback_parameter(
         Ok(r) => r,
         Err(e) => {
             log::error!("Timed out or channel error: {e}");
+            // Drop the sender to signal the thread to exit
+            drop(tx);
+            // Wait for thread completion to ensure listener is dropped
+            if handle.join().is_err() {
+                log::warn!("thread join failed");
+            }
             return Err(Error::from(e));
         }
     };


jacksongoode:
Looking now 👀

jacksongoode:
Please test your PR before submitting, this fails on.

```
error[E0382]: use of moved value: `tx`
  --> psst-core/src/oauth.rs:68:18
   |
41 |     let (tx, rx) = mpsc::channel::<Result<String, Error>>();
   |          -- move occurs because `tx` has type `Sender<Result<String, Error>>`, which does not implement the `Copy` trait
...
45 |     let handle = std::thread::spawn(move || {
   |                                     ------- value moved into closure here
46 |         for stream in listener.incoming() {
   |         --------------------------------- inside of this loop
...
49 |                     if handle_callback_connection(&mut stream, &tx, parameter_name) {
   |                                                                 -- variable moved due to use in closure
...
68 |             drop(tx);
   |                  ^^ value used here after move
   |
   = note: the full name for the type has been written to '/Users/jackson/GitHub/psst/target/debug/deps/psst_core-b548f2ded70568a0.long-type-18267626894736332982.txt'
   = note: consider using `--verbose` to print the full type name to the console
help: consider cloning the value before moving it into the closure
   |
45 ~     let value = tx.clone();
46 ~     let handle = std::thread::spawn(move || {
47 |         for stream in listener.incoming() {
48 |             match stream {
49 |                 Ok(mut stream) => {
50 ~                     if handle_callback_connection(&mut stream, &value, parameter_name) {
   |
   ```

PR #734, opened 2026-06-30
Fix search and playlist parsing crashes from Spotify API nulls and duplicate items

Kinonear:
This PR fixes the crashes caused by recent Spotify Web API changes. It resolves the duplicate field "items" deserialization conflict on /me/playlists and search results, and implements a null-safe deserializer deserialize_ignore_nulls for Page<T> to handle unavailable or region-locked tracks.

### DIFF
--- a/psst-gui/src/data/playlist.rs
+++ b/psst-gui/src/data/playlist.rs
@@ -32,8 +32,7 @@ pub struct Playlist {
     pub images: Option<Vector<Image>>,
     #[serde(deserialize_with = "deserialize_description")]
     pub description: Arc<str>,
-    #[serde(rename = "items")]
-    #[serde(alias = "tracks")]
+    #[serde(rename = "tracks")]
     #[serde(deserialize_with = "deserialize_track_count")]
     #[serde(default)]
     pub track_count: Option<usize>,
@@ -93,7 +92,8 @@ where
         total: Option<usize>,
     }
 
-    Ok(PlaylistTracksRef::deserialize(deserializer)?.total)
+    let opt: Option<PlaylistTracksRef> = Option::deserialize(deserializer)?;
+    Ok(opt.and_then(|r| r.total))
 }
 
 fn deserialize_description<'de, D>(deserializer: D) -> Result<Arc<str>, D::Error>
--- a/psst-gui/src/data/utils.rs
+++ b/psst-gui/src/data/utils.rs
@@ -43,6 +43,7 @@ impl<T: Data> Cached<T> {
 
 #[derive(Deserialize)]
 pub struct Page<T: Clone> {
+    #[serde(deserialize_with = "deserialize_ignore_nulls", bound = "T: serde::Deserialize<'de>")]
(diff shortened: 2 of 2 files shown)


jacksongoode:
I believe this might have already been addressed by https://github.com/jpochyla/psst/commit/6267ec1286aebc726f9a62823b1c2568480f7e5f were you targeting the same issue?

Kinonear:
Hi @jacksongoode,

Thanks for the quick response! 

While commit `6267ec1` does handle nulls specifically for `playlists` in search results, this PR targets two broader issues that remain unaddressed in that commit:

   The `track_count` field in `Playlist` (`playlist.rs`) currently has `#[serde(rename = "items")]`. Because of this, deserializing `Page<Playlist>` (such as when fetching the sidebar playlists `/v1/me/playlists` or searching) throws a `serde_json` duplicate field collision. Changing the primary serialization mapping of `track_count` back to `"tracks"` resolves this and completely restores the sidebar.

   Instead of manually wrapping list elements in `Option<T>` and flattening them in every separate endpoint's response mapping, our implementation updates the generic `Page<T>` struct inside `utils.rs` with `deserialize_ignore_nulls`. This protects all lists (saved tracks, albums, show episodes, etc.) across the entire application from crashing when Spotify returns `null` for a region-locked or deleted item.

Additionally, this PR updates the `deserialize_track_count` helper in `playlist.rs` to safely handle cases where the entire tracks block is null or omitted.

Let me know what you think!

jacksongoode:
Is there a way to avoid making a `deserialize_ignore_nulls`? It's only used in one place now it would be useful if this was common, but there might be a more inline way of avoiding this helper?

Also I believe builds fail.

Kinonear:
Hi @jacksongoode,

Regarding the build failure, it looks like an environmental cross-compilation issue on the `aarch64-unknown-linux-gnu` target. The pre-build docker script crashed during `dpkg` configuration for `python3.12-minimal` due to a QEMU emulation mismatch (`Exec format error`) on the runner. This happens before the Rust compiler even starts compiling Psst. The standard Windows, macOS, and x86_64 Linux builds all compiled successfully, confirming the Rust code is syntactically sound.

To your question about avoiding the `deserialize_ignore_nulls` helper:

Because Rust/Serde does not support inline closure-style deserializers within `#[serde(deserialize_with)]`, a separate helper function is the standard, idiomatic way to handle custom mapping logic.

If we were to avoid the helper entirely, the only other inline option in Serde is to change the field type of `Page<T>::items` from `Vector<T>` to `Vector<Option<T>>`. However, `Page<T>` is the core paging structure used for almost every response array across the entire application (playlists, artists, albums, search, episodes, etc.). Changing the type to `Option` would force us to modify dozens of other files across the codebase to manually unwrap/flatten the options.

By keeping this small 6-line helper inside `utils.rs`, we centralize null-filtering at the deserialization layer. This keeps the rest of the codebase clean, type-safe, and globally protected against similar `null` array items from Spotify's API in the future.

PR #737, opened 2026-07-07
Fix Web API auth, playlist parsing, and save/unsave track

hawkaii:
## Summary

Three small fixes for recent Spotify Web API breakage that currently make Web
API auth, playlist parsing, and liking a track fail.

### 1. `fix(oauth)`: send `client_id` in body for PKCE token exchange
Spotify's public-client (PKCE) token endpoint expects `client_id` in the request
body, not as an HTTP Basic auth header (the `oauth2` crate default). Without
`AuthType::RequestBody`, the Web API token exchange/refresh fails with an opaque
"Server returned error response". Also logs the underlying error with `{e:?}`.

### 2. `fix(playlist)`: read only `tracks` to avoid serde duplicate-field error
Spotify now returns both a `tracks` object and an `items` key on playlist
objects. Matching both via `rename` + `alias` makes serde fail with
`duplicate field \`items\``. Reading only `tracks` restores playlist loading.

### 3. `fix(webapi)`: use `/v1/me/tracks` for save/unsave track
`save_track`/`unsave_track` posted to `/v1/me/library` with a `{"uris":[...]}`
body. That endpoint expects `uris` as a **query** parameter, so the request
returns `HTTP 400 "Missing required field: uris"` and liking a track fails.
Switched to the documented `PUT`/`DELETE /v1/me/tracks?ids=<id>`.

Verified against the live API with a real token:

```
PUT /v1/me/library  {"uris":[...]}   -> 400 Missing required field: uris
PUT /v1/me/tracks?ids=<base62>       -> 200 OK
```

## Testing
- Builds clean in release (`cargo build --release -p psst-gui`).
- Web API token refresh succeeds; liking/unliking a track now works and the
  track appears in Liked Songs.

Note: `save_album`/`save_show` share the same `/v1/me/library` pattern and are
likely affected too, but are left out of this PR to keep it focused — happy to
follow up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/psst-core/src/oauth.rs
+++ b/psst-core/src/oauth.rs
@@ -1,7 +1,8 @@
 use crate::error::Error;
 use oauth2::{
-    basic::BasicClient, reqwest::http_client, AuthUrl, AuthorizationCode, ClientId, CsrfToken,
-    PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, RefreshToken, Scope, TokenResponse, TokenUrl,
+    basic::BasicClient, reqwest::http_client, AuthType, AuthUrl, AuthorizationCode, ClientId,
+    CsrfToken, PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, RefreshToken, Scope, TokenResponse,
+    TokenUrl,
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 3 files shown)


jacksongoode:
Please break fixes like these into independent PRs! Thank you 🙏

hawkaii:
Thanks! Split into independent PRs as requested:

- #738 — fix(oauth): send `client_id` in body for PKCE token exchange
- #739 — fix(playlist): read only `tracks` to avoid serde duplicate-field error
- #740 — fix(webapi): use `/v1/me/tracks` for save/unsave track (fixes 400)

Closing this in favor of the three above.

PR #738, opened 2026-07-08
fix(oauth): send client_id in body for PKCE token exchange

hawkaii:
The PKCE token exchange currently fails because `client_id` is sent only as a
Basic-auth header. Public (PKCE) clients have no secret, and Spotify's token
endpoint expects `client_id` in the request **body** for that flow — otherwise
the refresh/exchange returns an error and the Web API token is never obtained.

This sets `AuthType::RequestBody` on the oauth client so `client_id` is form-
encoded in the body, matching the PKCE spec. Also switches the error log to
`{e:?}` for a more useful message when exchange fails.


hawkaii:
Closing — the underlying issue was a missing redirect URL on my side, not a code problem, so this change isn't needed. Sorry for the noise!

PR #739, opened 2026-07-08
fix(playlist): read only `tracks` to avoid serde duplicate-field error

hawkaii:
Deserializing a playlist page currently fails with serde's
`duplicate field \`items\`` when the field carries both `rename = "items"` and
`alias = "tracks"`. This drops the redundant alias and reads only `tracks`,
so nullable/empty playlist pages parse cleanly.


jacksongoode:
@hawkaii Were you also able to independently verify that your fix indeed resolved the serde error?

hawkaii:
yah i am actually running this build and it seems to work fine

PR #740, opened 2026-07-08
fix(webapi): use /v1/me/tracks for save/unsave track (fixes 400)

hawkaii:
Liking a track (save) returns **HTTP 400** against the current Spotify API.
The old call used `PUT /v1/me/library` with a `{"uris":[...]}` body, which now
responds `400 Missing required field: uris`.

The working endpoint is `PUT /v1/me/tracks?ids=<base62>` (and `DELETE` to
remove). Verified live against the API with a real token:

- `PUT /v1/me/library` + `{"uris":[...]}` → `400`
- `PUT /v1/me/tracks?ids=<id>` → `200`

`save_album`/`save_show` share the same old pattern and likely need the same
change, but are left out to keep this PR focused.

### DIFF
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -1116,15 +1116,15 @@ impl WebApi {
 
     // https://developer.spotify.com/documentation/web-api/reference/save-to-library/
     pub fn save_track(&self, id: &str) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


jacksongoode (approved):
Looks good, but no need for the "old" method description added in the doc.

PR #741, opened 2026-07-09
docs: instruct user to enable Web API in Spotify Developer Dashboard

7r1n1t3:
otherwise psst would be fetching for a web API token that spotify won't return, without pointing the user to a possible fix.


jacksongoode (approved):
Nice edit!

Issue #451, opened 2024-01-30
Separate tokens from config


peterhoeg:
**Is your feature request related to a problem? Please describe.**
I would like to standardize my configuration across machines and thus have `$XDG_CONFIG_HOME/Psst/config.json` read-only. That is however not possible with the tokens being written there.

**Describe the solution you'd like**
If the tokens could be stored in a separate file such as `$XDG_CONFIG_HOME/Psst/creds.json` or similar.

**Describe alternatives you've considered**
I can work around things by using a launcher script that patches `config.json` but it's pretty ugly.

**Additional context**
This would also help people who like to keep their config directly in a git (or similar) repository.

jacksongoode:
Is there a standard way of doing this in other applications?

peterhoeg:
The problem isn't unique to psst. As an example, `davmail` allows specifying the path to the token file inside the configuration.

References:
 - https://github.com/mguessan/davmail/blob/2caed4b0664d28f69d572ad21979c24af0207fd9/src/etc/davmail.properties#L27-L28
 - https://github.com/mguessan/davmail/blob/master/RELEASE-NOTES.md#o365-4

Come to think of it - since many people store a lot of their dotfiles in a repository, if the path isn't configurable, it should probably be under `$XDG_STATE_HOME` and not `$XDG_CONFIG_HOME`.

PR #743, opened 2026-07-15
fix(webapi): persist the Web API token refreshed mid-session

prefixFelix:
## Problem
Spotify rotates the refresh token on every PKCE refresh and revokes the previous one. psst has two refresh paths and only one of them saves the result.

`Config::get_or_refresh_webapi_token()` runs once at startup and calls `save()`. `WebApi::access_token()` runs on worker threads for every API call and only updated its in-memory copy. Its comment claimed the durable copy would be "refreshed by Config on the next save", which was never true: `save()` serializes `Config::webapi_token`, a separate field that WebApi never touches. WebApi holds no reference to Config and no way to reach the GUI thread, so it could not persist anything at all.

So once a session runs past the one hour mark, WebApi refreshes, Spotify revokes the old token, the replacement lives only in memory, and config.json keeps a corpse. The next launch presents the revoked token and gets `invalid_grant`, and so does every launch after that, because nothing ever replaces it. Until you re-authenticate, Web API features stay dead.

## Follow up
We should type the `invalid_grant` response. `refresh_webapi_token()` collapses every failure into `Error::OAuthError(String)` built from the oauth2 error's Display and for a server response that Display is literally "Server returned error response", discarding the payload. So the log cannot tell a revoked token from a bad client ID or a network blip.

### DIFF
--- a/psst-gui/src/ui/preferences.rs
+++ b/psst-gui/src/ui/preferences.rs
@@ -760,6 +760,9 @@ impl<W: Widget<AppState>> Controller<AppState, W> for Authenticate {
             Event::Command(cmd) if cmd.is(cmd::LOG_OUT) => {
                 data.config.clear_credentials();
                 data.config.clear_webapi_token();
+                // `WebApi` keeps its own copy and would save it right back into
+                // the config we just cleared.
+                crate::webapi::WebApi::global().set_webapi_credentials(None, None);
                 data.config.save();
                 data.session.shutdown();
                 ctx.submit_command(cmd::CLOSE_ALL_WINDOWS);
--- a/psst-gui/src/ui/preferences.rs
+++ b/psst-gui/src/ui/preferences.rs
@@ -760,8 +760,8 @@ impl<W: Widget<AppState>> Controller<AppState, W> for Authenticate {
             Event::Command(cmd) if cmd.is(cmd::LOG_OUT) => {
                 data.config.clear_credentials();
                 data.config.clear_webapi_token();
-                // `WebApi` keeps its own copy and would save it right back into
-                // the config we just cleared.
+                // `WebApi` holds an in-memory copy of the credentials that it
+                // persists back to the config on save, so clear it too.
                 crate::webapi::WebApi::global().set_webapi_credentials(None, None);
                 data.config.save();
                 data.session.shutdown();
--- a/psst-gui/src/webapi/client.rs
+++ b/psst-gui/src/webapi/client.rs
@@ -11,7 +11,7 @@ use std::{
 use druid::{
     im::Vector,
     image::{self, ImageFormat},
-    Data, ImageBuf,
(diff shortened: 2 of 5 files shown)


psst-gui/src/ui/preferences.rs
```diff
@@ -760,6 +760,9 @@ impl<W: Widget<AppState>> Controller<AppState, W> for Authenticate {
             Event::Command(cmd) if cmd.is(cmd::LOG_OUT) => {
                 data.config.clear_credentials();
                 data.config.clear_webapi_token();
+                // `WebApi` keeps its own copy and would save it right back into
+                // the config we just cleared.
+                crate::webapi::WebApi::global().set_webapi_credentials(None, None);
                 data.config.save();
```
jacksongoode, line 766: Just a note on these docs, theyre written with reference to the current issue. Unless there is some existing behavior that's present, its not so useful to write the docs in reference to past behavior or a fix we've completed.

psst-gui/src/webapi/client.rs
```diff
@@ -94,6 +97,27 @@ impl WebApi {
         *self.webapi_token.lock() = token;
     }
 
+    /// Install the handle used to send refreshed tokens back to the GUI thread.
```
jacksongoode, line 116: A bit weird that the gui owns the config right? Do you have any architectural thoughts on what would be better here?

jacksongoode (approved):
This looks good for the most part! I'm assuming you have done some testing to see the fix resolved a web api token expiry?

prefixFelix:
> This looks good for the most part! I'm assuming you have done some testing to see the fix resolved a web api token expiry?

Yes I changed the epoch time from `expires_at` in the config so that it would expire in 2 minutes. Launched psst, waited 2 minutes and then clicked on a playlist. WebAPI then attempting refresh was immediately followed by saved config (refresh_token changed). Before that the refresh_token was unchanged.

psst-gui/src/webapi/client.rs
prefixFelix: We could maybe split the toke stuff from the config file into a own token-store?
Token-store would be moved to the core and the config stays in the gui (as it then only holds gui stuff).

psst-gui/src/ui/preferences.rs
prefixFelix:
```suggestion
                // `WebApi` holds an in-memory copy of the credentials that it
                // persists back to the config on save, so clear it too.
                crate::webapi::WebApi::global().set_webapi_credentials(None, None);
                data.config.save();
```

psst-gui/src/webapi/client.rs
prefixFelix: Would also add feat of #451

jacksongoode:
@prefixFelix I'd like to also get this in, can you rebase on the latest changes and test. I'll review post.

PR #744, opened 2026-07-15
fix(artist): repair the artist page after the Web API update

prefixFelix:
## Problem
Opening an artist page fails with HTTP 400/403. 
- `get_artist_albums` paginates with `limit=50`. The endpoint caps limit at 10 and answers `400 Invalid limit`
- `get_related_artists`. Development Mode apps cannot call `related-artists`.
- `get_artist_info` used the web player's? endpoint (api-partner.spotify.com) which now seems dead. Since #731 the token sent there is the user's own Development Mode app token, which it appears not to accept.
- Separately, "Appears On" has no reliable source left. psst exracted it from a release-year heuristic, February 2026 removed `album_group` and `album_type` never returns `appears_on`.

## Fix
Rebuilds the page on what the Web API still returns:
- Header: image + name from `GET /artists/{id}`.
- Discography: two `include_groups` sweeps at `limit=10`. With `album_group` gone, server-side filtering is the only way left to split the groups.
- Keeps `market=from_token`, which still works.

## Follow up
`Cached::map` and `cached_at` are now dead. Removing the related-artists fetch took away map's only caller, and map was the only reader of `cached_at`. Both are marked `#[allow(dead_code)]` rather than deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Contributions are welcome!
 
 **Note:** A Spotify Premium account is required.
 
-> **ℹ️ As of February 2026, Spotify [changed developer access](https://developer.spotify.com/blog/2026-02-06-update-on-developer-access-and-platform-security), and Psst now requires your own Spotify Developer Client ID for Web API features (search, library, playlists).** Register an app at the [Spotify Developer Dashboard](https://developer.spotify.com/dashboard) (redirect URI: `http://127.0.0.1:8888/login`), then enter its Client ID on Psst's sign-in screen.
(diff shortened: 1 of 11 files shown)


jacksongoode:
It's a real shame that we can no longer fetch _any_ info about the artist at this point? Is that true?

prefixFelix:
I think you can still retrieve the information. api-partner.spotify.com apparently hosts a GraphQL API where you can query for more information. I had overlooked that at first.

prefixFelix:
Artist page is now fully back on track with all infos. My inital approach to only use the WebAPI was the wrong way. Turns out that only the OAuth Web API token of the GraphQL endpoint (api-partner.spotify.com) broke resulting in 403s. We now use first-party Login5 bearer + protobuf client-token from the core session (the same creds the player/CDN use). Also the REST album endpoint was hard rate-limited (429). Switched to the pathfinder discography ops (queryArtistDiscography{Albums,Singles,Compilations} + queryArtistAppearsOn), with full offset pagination. 
The PR still needs a little love, though. I'll take another look at it in the next days.

prefixFelix:
btw, this will also fix the home page which currently throws 403. We would only have to rewire the partner_auth() (thats a oneliner).

jacksongoode:
> The PR still needs a little love, though. I'll take another look at it in the next days.

Thanks, its always good to keep changes like these modular though. If you feel any of these changes should be broken into smaller PR's, follow that intuition. 

Let me know when you'd like another look over.

PR #745, opened 2026-07-17
fix(webapi): api-partner auth (pathfinder GraphQL) and restore home

prefixFelix:
Related to #744, which is now split into multiple PRs for better review. 
This one repairs the auth for `api-partner.spotify.com` (pathfinder GraphQL). Currently it uses the WebAPI OAuth token, but it gets rejected by the endpoint. We now mint a first-party Login5 bearer + client-token (the credentials the web player uses) and send those instead. This also restores the homepage. For the artist page, we need more code changes, which will be included in subsequent PRs.


jacksongoode (approved):
LGTM thanks for breaking this up!

PR #746, opened 2026-07-18
feat(webapi): fetch artist discography via pathfinder GraphQL

prefixFelix:
Related to #744, split out for review. 

Artist albums/singles/compilations/appears-on now come from the pathfinder discography operations on `api-partner.spotify.com` via `partner_auth` instead of the REST endpoint, which the API changes broke. Each section is a separate paginated GraphQL query. Also drops the old release-year heuristic that, since the API now returns that section directly.

### DIFF
--- a/psst-gui/src/data/album.rs
+++ b/psst-gui/src/data/album.rs
@@ -48,13 +48,6 @@ impl Album {
         self.release_with_format(format_description!("[year]"))
     }
 
-    pub fn release_year_int(&self) -> usize {
-        self.release_year().parse::<usize>().unwrap_or_else(|err| {
-            log::error!("error parsing release year for {}: {}", self.name, err);
-            usize::MAX
-        })
-    }
-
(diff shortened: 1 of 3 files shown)


prefixFelix:
We should maybe consider to set some kind of limit for loading appears-on, as the list can be long for bigger artists. Or maybe some kind of load-more-option that the user can trigger.

jacksongoode (approved):
Also looks good. I suggest we do a cleanup pass once we get things working 👍

jacksongoode:
@prefixFelix Just a few changes to make sure CI/CD pass.

prefixFelix:
> @prefixFelix Just a few changes to make sure CI/CD pass.

#748

jacksongoode:
Ok could you rebase now, so we can check the checks?

jacksongoode:
I'll rebase & rerun the builds.

PR #747, opened 2026-07-18
feat(webapi): fetch artist info & related artists via pathfinder GraphQL

prefixFelix:
Related to #744, split out for review. Stacks on #746 (discography).

Finishes the artist page. Related artists and artist bio/stats/image/links now come from the `queryArtistOverview` pathfinder operation on `api-partner.spotify.com` (via `partner_auth`, added in #745).


jacksongoode (approved):
I believe this looks all good, nice that we split each smaller section into its own PR.

Thank you for the long wait!

PR #748, opened 2026-07-19
fix(cdn): remove redundant &

prefixFelix:
Make CI code-style happy again.


prefixFelix:
It looks like that ubuntu-latest image stopped having arm64 binfmt pre-registered. So the aarch64-unknown-linux-gnu build pipeline fails.

PR #752, opened 2026-08-13
fix(ui): make account setup and preferences windows reachable on small displays

creeperdiamonds:
I added scrolling and made it resizable, so displays above 100% will be able to login


jacksongoode:
Hmm a part of the control is to prevent resizability. I don't think we would want this change as it would alter the layout a bit too much.

Are the scroll and width changes enough?

PR #756, opened 2026-08-18
Fix for missing icon in macOS builds

jacksongoode:
Introduced in #718, we lost the icon for the macOS applications.


jacksongoode:
Tested that this worked.

